Issue #41, opened 2016-02-18
Add localization support for non moment words


manolo:
vaadin-locale-aware should consider other cases not only the moment library which only covers month and day names.

jounik:
First step:
- Provide a way for the developer to change all the strings/formats

Outside of 1.0 scope (not required, but nice to have):
- A generalized way to easily allow the developer to change the locale of all the vaadin elements in the app

PR #65, opened 2016-02-24
Implement localization via i18n object.

manolo:
Connected to #41

- user can customise messages using i18n object
- remove moment dependency
- fix infinite scroller to notify parent changes

[Review on Reviewable](https://reviewable.io/reviews/vaadin/vaadin-date-picker/65)


Saulis:
Review status: 0 of 15 files reviewed at latest revision, 4 unresolved discussions.

---

*[test/month-calendar.html, line 129 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65#-KBMR1KCnAeWTa_rSWmf:-KBMR1KCnAeWTa_rSWmg:1533426838) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/421b53e8e91754f4f346d0f1b7ca2c57294606e2/test/month-calendar.html#L129)):*
I think we oughta test also that mutation works: for example calling `datepicker.set('i18n.today', 'foobar');` and  verify that the text in today link has been updated.

---

*[vaadin-date-picker.html, line 147 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65#-KBMQRrML-uYhQZoSHGO:-KBMQRrN1Pe6MmIgL2bF:1400073284) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/421b53e8e91754f4f346d0f1b7ca2c57294606e2/vaadin-date-picker.html#L147)):*
I don't think we need `locale` property at this point yet since it doesn't do anything.

---

*[vaadin-date-picker.html, line 180 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65#-KBMPy4DfAXzBr0XKyki:-KBMPy4DfAXzBr0XKykj:37129328) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/421b53e8e91754f4f346d0f1b7ca2c57294606e2/vaadin-date-picker.html#L180)):*
I don't think we need to have an event at all, users will know that vaadin-date-picker can be localized so they can simply set datepicker.i18n to their liking.

---

*[vaadin-infinite-scroller.html, line 247 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65#-KBMPv4dLBfJUvJODBXt:-KBMPv4dLBfJUvJODBXu:-1007335549) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/421b53e8e91754f4f346d0f1b7ca2c57294606e2/vaadin-infinite-scroller.html#L247)):*
What is this change for?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65)*

tehapo:
Review status: 0 of 15 files reviewed at latest revision, 8 unresolved discussions.

---

*[locales/es.json, line 1 \[r2\]](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65#-KBMbzhUijeUu4nbOOjo:-KBMbzhVmTESfowSpt0h:1002961172) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/83ed37a477c65cc45714c89ad0d8cb01c8b88be7/locales/es.json#L1)):*
Is this file used anywhere?

---

*[locales/fi.json, line 1 \[r2\]](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65#-KBMc0Lktip_4kzMIAX1:-KBMc0Lktip_4kzMIAX2:1002961172) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/83ed37a477c65cc45714c89ad0d8cb01c8b88be7/locales/fi.json#L1)):*
Is this file used anywhere?

---

*[test/overlay.html, line 83 \[r2\]](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65#-KBM_HcymJT-MI5IQ5lM:-KBM_HcymJT-MI5IQ5lN:-1657940865) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/38f75f468805a7b1d7a4b237c54539b8d169a6e5/test/overlay.html#L83)):*
Why was this test removed?

---

*[vaadin-month-calendar.html, line 71 \[r2\]](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65#-KBMZzTtNl0ZMO99q69N:-KBMZzTudvWHrCgRfIDn:7332056) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/83ed37a477c65cc45714c89ad0d8cb01c8b88be7/vaadin-month-calendar.html#L71)):*
The `i18n` should be declared as a property.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65)*

tehapo:
Reviewed 14 of 15 files at r1, 1 of 1 files at r2.
Review status: all files reviewed at latest revision, 8 unresolved discussions.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65)*

manolo:
Review status: all files reviewed at latest revision, 8 unresolved discussions.

---

*[locales/es.json, line 1 \[r2\]](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65#-KBMbzhUijeUu4nbOOjo:-KBMg5Z_4pjAxG3ymMCX:585187596) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/83ed37a477c65cc45714c89ad0d8cb01c8b88be7/locales/es.json#L1)):*
This is tought to be used by the vaadin-i18n component. It loads component/locales/language.json and then visits all i18n.keys to localise them. Obviously this info should be provided by the component (vaadin-date-picker).

---

*[locales/fi.json, line 1 \[r2\]](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65#-KBMc0Lktip_4kzMIAX1:-KBMg_lgfCtdptdhaDMr:678021867) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/83ed37a477c65cc45714c89ad0d8cb01c8b88be7/locales/fi.json#L1)):*
same

---

*[test/month-calendar.html, line 129 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65#-KBMR1KCnAeWTa_rSWmf:-KBMjgD-syDu4Crlm75g:1779512369) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/421b53e8e91754f4f346d0f1b7ca2c57294606e2/test/month-calendar.html#L129)):*
actually these tests are in the vaadin-i18n-core prototype element, but right we have to add one here.

---

*[test/overlay.html, line 83 \[r2\]](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65#-KBM_HcymJT-MI5IQ5lM:-KBMhwYV2rdaEnzn9Lzp:594890253) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/38f75f468805a7b1d7a4b237c54539b8d169a6e5/test/overlay.html#L83)):*
It was moved to basic.html see 'should reflect value in overlay label' because formatter was duplicated in date picker and overlay.

---

*[vaadin-date-picker.html, line 147 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65#-KBMQRrML-uYhQZoSHGO:-KBMiMKtQHssL11z3rUQ:-1604690902) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/421b53e8e91754f4f346d0f1b7ca2c57294606e2/vaadin-date-picker.html#L147)):*
locale in the component is thought to override locale defined in a global component, the idea is this:

```
<vaadin-i18n-core locale="es"></vaadin-i18n-core>
<vaadin-date-picker label="Cumpleaños"></vaadin-date-picker>
<another-vaadin-component></another-vaadin-component>
<vaadin-date-picker label="syntymäpäivä" locale="fi"></vaadin-date-picker>
```

So as all the UI is in Spanish but the last date picker

---

*[vaadin-date-picker.html, line 180 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65#-KBMPy4DfAXzBr0XKyki:-KBMip7q_lgbNmZM_2T4:-771085561) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/421b53e8e91754f4f346d0f1b7ca2c57294606e2/vaadin-date-picker.html#L180)):*
This is for notifying  <vaadin-i18n-component>  to internationalise this component, it's optional that the user includes it in the UI though.

---

*[vaadin-infinite-scroller.html, line 247 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65#-KBMPv4dLBfJUvJODBXt:-KBMj4sZirKMwUU2xxZc:297864945) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/421b53e8e91754f4f346d0f1b7ca2c57294606e2/vaadin-infinite-scroller.html#L247)):*
when you call datepicker.set('i18n.foo', 'bar')  or notifyPath  templatizer is not notified otherwise

---

*[vaadin-month-calendar.html, line 71 \[r2\]](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65#-KBMZzTtNl0ZMO99q69N:-KBMjRDE4Q5WZeTUoWRi:-483562451) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/83ed37a477c65cc45714c89ad0d8cb01c8b88be7/vaadin-month-calendar.html#L71)):*
it's not mandatory to declare a property if you don't want a public api documentation.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65)*

manolo:
Review status: all files reviewed at latest revision, 8 unresolved discussions.

---

*[vaadin-date-picker.html, line 180 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65#-KBMPy4DfAXzBr0XKyki:-KBMl_r3zgWcjQqTShkD:1636626347) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/421b53e8e91754f4f346d0f1b7ca2c57294606e2/vaadin-date-picker.html#L180)):*
I meant notify the vaadin-i18n-core component, which visit i18n properties to internationalize it based on the locale. 
The reason why I use an event here is to decouple vaadin-date-picker or vaadin-uploader from any i18n framework, the user can implement it's own i18n solution listening for this event, or just use the out-of-the-box solution provided by vaadin-i18n-core element

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65)*

samiheikki:
Reviewed 14 of 15 files at r1, 1 of 1 files at r2.
Review status: all files reviewed at latest revision, 10 unresolved discussions.

---

*[demo/index.html, line 34 \[r2\]](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65#-KBMewyOJ5TuL1vij7Jl:-KBMewyPTPwTjZWrg3ZC:593450321) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/83ed37a477c65cc45714c89ad0d8cb01c8b88be7/demo/index.html#L34)):*
Can monthNames be created same way in both examples? With `split` or with `array`.

---

*[vaadin-date-picker.html, line 154 \[r2\]](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65#-KBMkVCbFjzT5fx8nb4n:-KBMkVCc8ZEfec784U1O:-498613107) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/83ed37a477c65cc45714c89ad0d8cb01c8b88be7/vaadin-date-picker.html#L154)):*
localize with z?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65)*

manolo:
Review status: all files reviewed at latest revision, 10 unresolved discussions.

---

*[demo/index.html, line 34 \[r2\]](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65#-KBMewyOJ5TuL1vij7Jl:-KBMo_yiBHg5EQPWPru0:1999976460) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/83ed37a477c65cc45714c89ad0d8cb01c8b88be7/demo/index.html#L34)):*
Well, that is deliberated because I wanted to show both cases to the user.
The split way could be confusing for some users, so in case of unifying I'd rather the array.

---

*[vaadin-date-picker.html, line 154 \[r2\]](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65#-KBMkVCbFjzT5fx8nb4n:-KBMp0RMtOZ36ftDQFIx:-379181850) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/83ed37a477c65cc45714c89ad0d8cb01c8b88be7/vaadin-date-picker.html#L154)):*
Depends on the type of english you use es-US(z) en-GB(s)

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65)*

manolo:
Review status: all files reviewed at latest revision, 10 unresolved discussions.

---

*[vaadin-date-picker.html, line 154 \[r2\]](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65#-KBMkVCbFjzT5fx8nb4n:-KBMpOcrHtkCoGjeLwnV:-489684638) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/83ed37a477c65cc45714c89ad0d8cb01c8b88be7/vaadin-date-picker.html#L154)):*
http://forum.wordreference.com/threads/localized-vs-localised-software.1315905/

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65)*

manolo:
Review status: 13 of 15 files reviewed at latest revision, 8 unresolved discussions.

---

*[test/month-calendar.html, line 129 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65#-KBMR1KCnAeWTa_rSWmf:-KBNRLxlu11R_sZq-Jkw:1188050664) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/421b53e8e91754f4f346d0f1b7ca2c57294606e2/test/month-calendar.html#L129)):*
Added a couple of tests for checking that mutation is notified correctly to children

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65)*

tehapo:
Review status: 13 of 15 files reviewed at latest revision, 9 unresolved discussions, some commit checks broke.

---

*[vaadin-date-picker.html, line 154 \[r2\]](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65#-KBMkVCbFjzT5fx8nb4n:-KBNmMQrm7KQlkWXqWr2:-48627958) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/83ed37a477c65cc45714c89ad0d8cb01c8b88be7/vaadin-date-picker.html#L154)):*
I think we have so far been always going with the en-US variations on other places.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65)*

manolo:
Review status: 13 of 15 files reviewed at latest revision, 9 unresolved discussions.

---

*[vaadin-date-picker.html, line 154 \[r2\]](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65#-KBMkVCbFjzT5fx8nb4n:-KBNsqmzYZAWimuSLUZx:-499146009) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/83ed37a477c65cc45714c89ad0d8cb01c8b88be7/vaadin-date-picker.html#L154)):*
Done.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65)*

platosha:
Review status: 13 of 15 files reviewed at latest revision, 12 unresolved discussions.

---

*[vaadin-date-picker.html, line 169 \[r4\]](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65#-KBO6YMHuOaJAcZ1z9a2:-KBO6YMHuOaJAcZ1z9a3:1917049610) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/52d688339aafb6bfa8d1fa3ef6fab109777085c3/vaadin-date-picker.html#L169)):*
Just for discussion of possible improvement with the team: would it be better to not use a fixed US-style format by default, but call `date.toLocaleDateString()` instead?

Of course, this method doesn't provide cross-browser support for localized date formats, so we can't use it with the locale specified by the `locale` property. But still it can format a nicely localised date string to the user's browser/system locale.

---

*[vaadin-date-picker.html, line 185 \[r4\]](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65#-KBO4TIHA8lgIqejRN3k:-KBO4TIIE674TF_P8nS-:-2137328721) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/52d688339aafb6bfa8d1fa3ef6fab109777085c3/vaadin-date-picker.html#L185)):*
This method is used in computed bindigs, and its outcome depends on `i18n.formatDate` value. Would make sense to add `formatDate` as a second parameter to indicate that.

---

*[vaadin-month-calendar.html, line 73 \[r4\]](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65#-KBOB2ZSJmqTrh9d5W_E:-KBOB2ZSJmqTrh9d5W_F:1542232230) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/52d688339aafb6bfa8d1fa3ef6fab109777085c3/vaadin-month-calendar.html#L73)):*
Why add `i18n` as a separate dependency? Shouldn't it update the values with just a sub-property dependency?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65)*

tomivirkki:
*[vaadin-date-picker.html, line 185 \[r4\]](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65#-KBO4TIHA8lgIqejRN3k:-KBg0L3TOfJHo5Qfzl5n:-181833436) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/52d688339aafb6bfa8d1fa3ef6fab109777085c3/vaadin-date-picker.html#L185)):*
I don't like the idea that vaadin-date-picker itself is responsible of formatting a date based on a provided pattern. That should be the job of a utility like moment.js.

i18n.formatDate should be a function which vaadin-date-picker uses to get a formatted date so the component itself would have no say which utility is used.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65)*

tomivirkki:
Review status: 13 of 15 files reviewed at latest revision, 13 unresolved discussions, some commit checks failed.

---

*[vaadin-date-picker.html, line 180 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65#-KBMPy4DfAXzBr0XKyki:-KBg3GjHa-5jQAVsFB03:-509182362) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/421b53e8e91754f4f346d0f1b7ca2c57294606e2/vaadin-date-picker.html#L180)):*
This functionality shouldn't be here as vaadin-i18n-core (or whatever we decide to name it) doesn't exists yet.

---

*[vaadin-date-picker.html, line 157 \[r4\]](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65#-KBg2tvmQIIdQKoW0aHo:-KBg2tvmQIIdQKoW0aHp:691707205) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/52d688339aafb6bfa8d1fa3ef6fab109777085c3/vaadin-date-picker.html#L157)):*
'internationalizing' and 'customized'. vaadin-i18n-core doesn't exists so it shouldn't be mentioned here.

---

*[vaadin-month-calendar.html, line 71 \[r2\]](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65#-KBMZzTtNl0ZMO99q69N:-KBg1URZ2gUIw4fIZeFz:1048825524) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/83ed37a477c65cc45714c89ad0d8cb01c8b88be7/vaadin-month-calendar.html#L71)):*
Why shouldn't it be in the public api of vaadin-month-calendar? It is an independent module after all

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65)*

tomivirkki:
*[vaadin-date-picker.html, line 18 \[r4\]](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65#-KBg40dGtWXBwe0NxocY:-KBg40dHpk8tpoSGFBcu:1485344313) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/52d688339aafb6bfa8d1fa3ef6fab109777085c3/vaadin-date-picker.html#L18)):*
'The user' refers to the developer in this case...a bit confusing. Perhaps it should be left out and say "...This element can be internationalized by..."

Not sure this is the right place to have a code example for the functionality. After all, i18n isn't specific to this component but other components will share the same API as well.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65)*

manolo:
Review status: 13 of 15 files reviewed at latest revision, 14 unresolved discussions, some commit checks failed.

---

*[vaadin-date-picker.html, line 180 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65#-KBMPy4DfAXzBr0XKyki:-KBqpnDmTqLmANnpqSKY:35813615) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/421b53e8e91754f4f346d0f1b7ca2c57294606e2/vaadin-date-picker.html#L180)):*
Disagree, I think we can document this event, so as the user can do it's own implementation.
It may be a different name, like vaadin-element-ready, and then not only i18n but other components could take advantage of it. But definitively I think we should leave and event here, instead of updating the API in the future.

---

*[vaadin-date-picker.html, line 18 \[r4\]](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65#-KBg40dGtWXBwe0NxocY:-KBqc50rFFBDaNbYPRs_:-836915838) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/52d688339aafb6bfa8d1fa3ef6fab109777085c3/vaadin-date-picker.html#L18)):*
Agree removing this from here. I think it should be documented in the asciidoc though.

---

*[vaadin-date-picker.html, line 157 \[r4\]](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65#-KBg2tvmQIIdQKoW0aHo:-KBqsTiDIDQH3HYPmebF:-37812484) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/52d688339aafb6bfa8d1fa3ef6fab109777085c3/vaadin-date-picker.html#L157)):*
agree, we have to remove this, at least in the meanwhile that element does not exist

---

*[vaadin-date-picker.html, line 169 \[r4\]](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65#-KBO6YMHuOaJAcZ1z9a2:-KBqp5G9oi-8IUdyb5oa:-1266686492) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/52d688339aafb6bfa8d1fa3ef6fab109777085c3/vaadin-date-picker.html#L169)):*
IMO, that has more inconvenient than advantage:
-  every browser returns a different string
- the developer would want a fixed language, for instance english
- there is not information to internationalise month and day names in the calendar
- so the user seeing spanish month names in the input and english in the overlay would be a bad experience.

---

*[vaadin-date-picker.html, line 185 \[r4\]](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65#-KBO4TIHA8lgIqejRN3k:-KBqqNtxdcS1786ONUQa:1284292896) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/52d688339aafb6bfa8d1fa3ef6fab109777085c3/vaadin-date-picker.html#L185)):*
Yes, the second parameter would be ok. 
I think vaadin-date-picker should give a basic functionality of formatting and parsing dates, its a date element so the developer would expect that feature out-of-the-box, otherwise he has to include something else always. I have studied what JS formatting libraries there are and none is good enough because they use non-standard conventions, or they use other dependencies, or they are difficult to find.
Vaadin-date-picker should have a minimum friction for the user to implement a custom format, but the user has the option to improve parsers and formatters just overriding the fucntion

---

*[vaadin-month-calendar.html, line 71 \[r2\]](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65#-KBMZzTtNl0ZMO99q69N:-KBqrRaPByATM_Xf0v3o:485694650) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/83ed37a477c65cc45714c89ad0d8cb01c8b88be7/vaadin-month-calendar.html#L71)):*
I didn't want to maintain an API documentation if not absolutely necessary. Since vaadin-date-picker could write this property arbitrary, I'd rather not cause other issues if the user already set it.

---

*[vaadin-month-calendar.html, line 73 \[r4\]](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65#-KBOB2ZSJmqTrh9d5W_E:-KBqsJkZZF2BdrHpe-Rw:-933465605) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/52d688339aafb6bfa8d1fa3ef6fab109777085c3/vaadin-month-calendar.html#L73)):*
what do you mean?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65)*

tomivirkki:
This implementation makes many assumptions on the i18n API that hasn't even been planned/decided yet. Before we have such a thing in the roadmap, single components should not make any premature guesses in their apis regarding it.

I'm talking about "vaadin-i18n-register" event, "locale" property and mentions of "<vaadin-i18n-core>" all over.

At this point we only need "datepicker.i18n" object without anything extra. Only once we know where to go from there we can make the necessary changes.

---

Review status: 13 of 15 files reviewed at latest revision, 14 unresolved discussions, some commit checks failed.

---

*[vaadin-date-picker.html, line 185 \[r4\]](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65#-KBO4TIHA8lgIqejRN3k:-KBrTl2s7bVJrZzmoymM:-1701018182) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/52d688339aafb6bfa8d1fa3ef6fab109777085c3/vaadin-date-picker.html#L185)):*
What second parameter?

"I think vaadin-date-picker should give a basic functionality of formatting and parsing dates, its a date element so the developer would expect that feature out-of-the-box"
Then why remove moment.js in the first place if we want to help as much as possible? Why offer date formatting BUT NOT ready translations...?

"I have studied what JS formatting libraries there are and none is good enough "
So you say we should make a better one? Inside vaadin-date-picker?

My original point was to require i18n.formatDate to be a _function_. So the responsibility of formatting is transferred to the client (just like this PR transfers the responsibility of firstDayOfWeek determination, monthNames... to the client)

Then what we'd do in vaadin-date-picker would be something like "var formatted = this.i18n.formatDate(value);".

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65#-:-KBrQeKVFhfS2UBMl2uE:-1427594127)*

manolo:
Well, this implementation is previous to the decision of whether we needed a discussion about i18n. The only information in cards were that we needed to internationalise a button in the UI. So starting at this point, I came with a simple solution able to handle any cases, the user internationalising elements via api, or anyone creating external stuff to internationalise after the component was created. There were already a couple of chats in skype about that.

The event can be renamed to vaadin-element-register or whatever to decouple from the idea of i18n, but from my point of view it's important that or components notify that they are in the dom so as other elements could interact with them, for instance this could help with some tests waiting for components to be ready. But insist if we don't have this event from the beginning prototypes are hard to do because we have to patch all components with this basic thing, I don't see any big reason for delaying this, is so trivial to me that a discussion about the convenience or not is a waste of time. Probably if you suspect that the API which basically is that the event name changes is going to change, we can just leave this undocumented, or with a notice.

Anyway, if it does not make any sense to  you to leave this event here I'll remove with pleasure.

---

Review status: 13 of 15 files reviewed at latest revision, 14 unresolved discussions, some commit checks failed.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65#-:-KBvME4fdnt0ttysk3mV:1184526525)*

manolo:
Review status: 13 of 15 files reviewed at latest revision, 14 unresolved discussions, some commit checks failed.

---

*[vaadin-date-picker.html, line 185 \[r4\]](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65#-KBO4TIHA8lgIqejRN3k:-KBvOevtGzNNAneSrH1P:-44788329) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/52d688339aafb6bfa8d1fa3ef6fab109777085c3/vaadin-date-picker.html#L185)):*
> Then why remove moment.js in the first place if we want to help as much as possible? Why offer date formatting BUT NOT ready translations...?

Because you force to the user to a hardcoded dependency of moment although he wants to replace it by something better like cldr.

> So you say we should make a better one? Inside vaadin-date-picker?

I meant that none of them is light enough to depend on, and light examples I saw were using non standard syntax (but php, etc) . We need some kind of formatting though, I insist, not a complete feature, but a limited one in order not to bloat the component.

> My original point was to require i18n.formatDate to be a function.

In this PR,  if i18n.formatDate == function we use it, so as the user has full control on formatting. So, nothing changes in your proposal, except that when it's not a function we use our lite formatter. In this case we offer an out-of-the-box solution for users not demanding complex formatters (I guess a hight percent).

> something like "var formatted = this.i18n.formatDate(value);".

That is what this._formatValue(value) does. 

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65)*

tomivirkki:
Review status: 13 of 15 files reviewed at latest revision, 14 unresolved discussions, some commit checks failed.

---

*[vaadin-date-picker.html, line 185 \[r4\]](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65#-KBO4TIHA8lgIqejRN3k:-KBvZcNV1YuYYEZMZ-d3:-415919802) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/52d688339aafb6bfa8d1fa3ef6fab109777085c3/vaadin-date-picker.html#L185)):*
Didn't notice that it accepts a function as well. Then it's much more flexible.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65)*

Saulis:
Have to agree with Tomi here, I would keep the API changes to absolute minimum until we know more about the Polymer i18n support. It's easier to add more stuff later in 1.x versions than to remove if we guess something wrong.

The `i18n` property provides the simplest possible way of overriding the strings in `vaadin-date-picker` and `vaadin-upload`, yet its still lightweight enough solution to be removed or replaced later.

As I've commented below, I don't really fully get the idea behind firing the custom event. If it's about knowing when the elements are upgraded before setting the `i18n` object, it's not needed anymore in Polymer 1.3.0.

---

Review status: 13 of 15 files reviewed at latest revision, 13 unresolved discussions, some commit checks failed.

---

*[vaadin-date-picker.html, line 147 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65#-KBMQRrML-uYhQZoSHGO:-KBwC6zaWBOWTZa7KQdj:-331540622) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/421b53e8e91754f4f346d0f1b7ca2c57294606e2/vaadin-date-picker.html#L147)):*
Well right, but `_localeChanged` doesn't do anything but fire an event `vaadin-i18n-register` which I feel is redundant. I would want to keep API changes to a minimum until we know more about the upcoming i18n support in Polymer.

---

*[vaadin-date-picker.html, line 180 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65#-KBMPy4DfAXzBr0XKyki:-KBwBftDFn2b1oeN3RCa:-1416572199) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/421b53e8e91754f4f346d0f1b7ca2c57294606e2/vaadin-date-picker.html#L180)):*
I still don't understand what is the point of this event. Could you provide a full example how the handler could be implemented?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65#-:-KBwDPGH82GD6E_O5pfb:-1503410850)*

manolo:
> If it's about knowing when the elements are upgraded before setting the i18n object, it's not needed anymore in Polymer 1.3.0.

It's not a matter of when the element is ready, but a pluggable solution localising vaadin components which is absolutly decoupled. It should be a pin if that component have to transverse the dom to figure out which components have a i18n object to update, or observe dom mutations for new elements coming to the ui that could be eventually internationalised.

---

Review status: 13 of 15 files reviewed at latest revision, 13 unresolved discussions, some commit checks failed.

---

*[vaadin-date-picker.html, line 147 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65#-KBMQRrML-uYhQZoSHGO:-KBwW2vOKuXsMZhuZKba:-1568835748) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/421b53e8e91754f4f346d0f1b7ca2c57294606e2/vaadin-date-picker.html#L147)):*
That is to inform an external component that this specific element in the UI wants to change its locale. It's a very simple implementation.  Otherwise it should be almost impossible to have two date pickers in the same UI with different locales.

---

*[vaadin-date-picker.html, line 180 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65#-KBMPy4DfAXzBr0XKyki:-KBwWNk-VO6_LfaFMp37:2129497902) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/421b53e8e91754f4f346d0f1b7ca2c57294606e2/vaadin-date-picker.html#L180)):*
explained above

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65#-:-KBwVIljwSBJAafbJMUg:1281914221)*

manolo:
Review status: 4 of 13 files reviewed at latest revision, 13 unresolved discussions, some commit checks failed.

---

*[vaadin-date-picker.html, line 185 \[r4\]](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65#-KBO4TIHA8lgIqejRN3k:-KC0PejG49_HWkDEh7nE:964880097) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/52d688339aafb6bfa8d1fa3ef6fab109777085c3/vaadin-date-picker.html#L185)):*
in latest I removed that

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65)*

Saulis:
Review status: 1 of 12 files reviewed at latest revision, 9 unresolved discussions.

---

*[vaadin-month-calendar.html, line 73 \[r4\]](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65#-KBOB2ZSJmqTrh9d5W_E:-KC0XDPxBFJQl3xZROXv:-406649656) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/52d688339aafb6bfa8d1fa3ef6fab109777085c3/vaadin-month-calendar.html#L73)):*
The `i18n` parameter is not used in the function `_getWeekDayNames`

---

*[demo/index.html, line 50 \[r9\]](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65#-KC0YEtldSIjR0kZ78C3:-KC0YEtldSIjR0kZ78C4:538964573) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/7c892948bebf78ba8d5a26e4e4b09a2a418e3e14/demo/index.html#L50)):*
How about `Localising` or `Localizing` how ever it's spelled?

---

*[demo/index.html, line 56 \[r10\]](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65#-KC0YumG9SzgzehZraV_:-KC0YumG9SzgzehZraVa:1210931244) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/4dbfa308d28efc84dfcaf0a1710b607b49ee844b/demo/index.html#L56)):*
Remove `HTMLImport.whenReady`

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65)*

manolo:
Review status: 1 of 12 files reviewed at latest revision, 9 unresolved discussions.

---

*[vaadin-month-calendar.html, line 73 \[r4\]](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65#-KBOB2ZSJmqTrh9d5W_E:-KC0ZY5Z0ivMQknlcL3E:-1296469852) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/52d688339aafb6bfa8d1fa3ef6fab109777085c3/vaadin-month-calendar.html#L73)):*
That's is there to observe i18n, otherwise only i18n.weekdaysShort and i18n.firstDayOfWeek would change the items, but not e.i18n = '{}'

---

*[demo/index.html, line 50 \[r9\]](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65#-KC0YEtldSIjR0kZ78C3:-KC0_FBkeiMbr6rTRIvx:-499146009) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/7c892948bebf78ba8d5a26e4e4b09a2a418e3e14/demo/index.html#L50)):*
Done.

---

*[demo/index.html, line 56 \[r10\]](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65#-KC0YumG9SzgzehZraV_:-KC0_Er_-S1J4ll-g8d4:-499146009) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/4dbfa308d28efc84dfcaf0a1710b607b49ee844b/demo/index.html#L56)):*
Done.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65)*

Saulis:
Added comments about the `formatted` binding, otherwise :lgtm:

---

Review status: 1 of 12 files reviewed at latest revision, 12 unresolved discussions.

---

*[vaadin-date-picker-overlay.html, line 242 \[r10\]](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65#-KC0aq0QdbneMopyLyP5:-KC0aq0RsmJaxjCYPlQL:1121913249) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/4dbfa308d28efc84dfcaf0a1710b607b49ee844b/vaadin-date-picker-overlay.html#L242)):*
Change this to `_formattedDisplayValue` and declare it in the `properties` 

---

*[vaadin-date-picker.html, line 150 \[r10\]](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65#-KC0aSAy8fmGYyE6PR_j:-KC0aSAy8fmGYyE6PR_k:-817918983) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/4dbfa308d28efc84dfcaf0a1710b607b49ee844b/vaadin-date-picker.html#L150)):*
Rather bind to `_formattedDisplayValue="[[_formatDisplayValue(_selectedDate, i18n.formatDate)]]"

---

*[vaadin-date-picker.html, line 301 \[r10\]](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65#-KC0aGWaX5RHR5zxtEme:-KC0aGWaX5RHR5zxtEmf:-26020936) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/4dbfa308d28efc84dfcaf0a1710b607b49ee844b/vaadin-date-picker.html#L301)):*
I'd rename this to `_formatDisplayValue` to be more explicit. Also, should just return the formatted value instead of assigning it to `this.formatted` to avoid side-effects.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65#-:-KC0awzesTuLRsc9cmxV:351961260)*

tomivirkki:
Review status: 1 of 12 files reviewed at latest revision, 16 unresolved discussions.

---

*[vaadin-date-picker.html, line 304 \[r10\]](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65#-KC0b_4LwhAbAIow_qZg:-KC0b_4LwhAbAIow_qZh:-599691047) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/4dbfa308d28efc84dfcaf0a1710b607b49ee844b/vaadin-date-picker.html#L304)):*
== -> ===
When can the 'value' be a string?

---

*[vaadin-month-calendar.html, line 116 \[r10\]](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65#-KC0avvkqwjiRIaEoJo6:-KC0avvkqwjiRIaEoJo7:225210013) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/4dbfa308d28efc84dfcaf0a1710b607b49ee844b/vaadin-month-calendar.html#L116)):*
Change to i18n: Object

---

*[demo/index.html, line 63 \[r10\]](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65#-KC0ZlU_XtdJrqp-CbfK:-KC0ZlUa6WyBfAexyEQq:616048488) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/4dbfa308d28efc84dfcaf0a1710b607b49ee844b/demo/index.html#L63)):*
Why not inline arrays?

---

*[test/basic.html, line 261 \[r10\]](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65#-KC0aUv13434n5kcCUQ0:-KC0aUv2VcjlbD1LccI4:816121469) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/4dbfa308d28efc84dfcaf0a1710b607b49ee844b/test/basic.html#L261)):*
You use an inline array 18 rows below so why not here?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65)*

manolo:
Review status: 1 of 12 files reviewed at latest revision, 16 unresolved discussions.

---

*[vaadin-date-picker.html, line 304 \[r10\]](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65#-KC0b_4LwhAbAIow_qZg:-KC0fMIrXEtpdN8ZuzQA:-1066862908) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/4dbfa308d28efc84dfcaf0a1710b607b49ee844b/vaadin-date-picker.html#L304)):*
== and === works the same when comparing with a constant string, only if I was comparing with empty string should make sense to use the ===

It was necessary when I used to pass the value, now there is a new property which is _selectedDate

---

*[demo/index.html, line 63 \[r10\]](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65#-KC0ZlU_XtdJrqp-CbfK:-KC0hWygmp_Ibt776lvW:1396232473) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/4dbfa308d28efc84dfcaf0a1710b607b49ee844b/demo/index.html#L63)):*
less code and lazy. I copied this from moment

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65)*

platosha:
:lgtm_strong: :+1:

---

Review status: 1 of 12 files reviewed at latest revision, 16 unresolved discussions.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65#-:-KC0jq6_1_WpwkubTZBx:88293517)*

manolo:
Review status: 1 of 12 files reviewed at latest revision, 16 unresolved discussions.

---

*[vaadin-date-picker-overlay.html, line 242 \[r10\]](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65#-KC0aq0QdbneMopyLyP5:-KC0lTBUZxf_l-GMJAsI:-602575704) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/4dbfa308d28efc84dfcaf0a1710b607b49ee844b/vaadin-date-picker-overlay.html#L242)):*
So how do I name the attribute used by the parent to set this?
_formated-display-value ? I think that is not readable

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/65)*

PR #78, opened 2016-02-29
Add arrow key and enter handlers.

### LINKED ISSUE #76: Keyboard navigation
This feature is required before we can start working on ARIA support #70

https://www.w3.org/TR/wai-aria-practices/#datepicker

Saulis:
Fixes #76

[Review on Reviewable](https://reviewable.io/reviews/vaadin/vaadin-date-picker/78)

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -33,7 +33,8 @@
     "paper-styles": "polymerelements/paper-styles#^1.1.3",
     "moment": "^2.11.2",
     "iron-media-query": "polymerelements/iron-media-query#^1.0.8",
-    "paper-ripple": "polymerelements/paper-ripple#^1.0.5"
+    "paper-ripple": "polymerelements/paper-ripple#^1.0.5",
+    "iron-a11y-keys-behavior": "polymerelements/iron-a11y-keys-behavior#^1.1.1"
   },
   "devDependencies": {
     "web-component-tester": "^4.1.0",
--- a/bower.json
+++ b/bower.json
@@ -38,7 +38,7 @@
   },
   "devDependencies": {
     "web-component-tester": "^4.1.0",
-    "iron-test-helpers": "polymerelements/iron-test-helpers#^1.0.0",
+    "iron-test-helpers": "polymerelements/iron-test-helpers#^1.1.5",
     "iron-component-page": "polymerelements/iron-component-page#~1.1.0",
     "iron-form": "PolymerElements/iron-form#^1.0.15",
     "iron-demo-helpers": "polymerelements/iron-demo-helpers#^1.1.1",
--- a/docs/vaadin-date-picker-overview.adoc
+++ b/docs/vaadin-date-picker-overview.adoc
@@ -20,6 +20,22 @@ The dropdown footer has a today button for quickly jumping to the current month.
 It does not select a date but only scrolls the calendar view.
 A cancel button in the footer allows closing the dropdown without affecting the value.
 
+== Keyboard Navigation Overview
+When a date is focused using the keyboard, it can be selected or deselected using Space key. +
+With Enter key, the focused date is selected and the date picker is closed.
+
+Focus can be moved with the following keys: +
+Up: a week backward +
+Down: a week forward +
+Left: a day backward +
+Right: a day forward +
+Home: focus first day of the month +
+End: focus last day of the month +
+PageUp: a month backward +
+PageDown: a month forward +
+Shift+PageUp: a year backward +
+Shift+PageDown: a year forward
+
 The current alpha version has a limited set of features.
 Future versions will address issues such as date ranges and keyboard input.
 
--- a/test/common.js
+++ b/test/common.js
@@ -1,3 +1,5 @@
+var ios = /iPad|iPhone|iPod/.test(navigator.userAgent) && !window.MSStream;
+
 function tap(element) {
   Polymer.Base.fire('tap', {}, {
     bubbles: true,
--- a/test/index.html
+++ b/test/index.html
@@ -16,7 +16,8 @@
       'overlay.html',
       'month-calendar.html',
       'scroller.html',
-      'form-input.html'
+      'form-input.html',
+      'keyboard-navigation.html'
     ]);
   </script>
 </body>
--- a/test/keyboard-navigation.html
+++ b/test/keyboard-navigation.html
@@ -27,52 +27,60 @@
       describe('keyboard navigation with arrow keys', function() {
         var datepicker;
 
-        function arrowDown() {
-          MockInteractions.keyDownOn(getInput(), 40);
+        function arrowDown(target) {
+          MockInteractions.keyDownOn(target || overlay(), 40);
         }
 
         function arrowRight() {
-          MockInteractions.keyDownOn(getInput(), 39);
+          MockInteractions.keyDownOn(overlay(), 39);
         }
 
-        function arrowUp() {
-          MockInteractions.keyDownOn(getInput(), 38);
+        function arrowUp(target) {
+          MockInteractions.keyDownOn(target || overlay(), 38);
         }
 
         function arrowLeft() {
-          MockInteractions.keyDownOn(getInput(), 37);
+          MockInteractions.keyDownOn(overlay(), 37);
         }
 
         function home() {
-          MockInteractions.keyDownOn(getInput(), 36);
+          MockInteractions.keyDownOn(overlay(), 36);
         }
 
         function end() {
-          MockInteractions.keyDownOn(getInput(), 35);
+          MockInteractions.keyDownOn(overlay(), 35);
         }
 
         function pageDown(modifiers) {
-          MockInteractions.keyDownOn(getInput(), 34, modifiers);
+          MockInteractions.keyDownOn(overlay(), 34, modifiers);
         }
 
         function pageUp(modifiers) {
-          MockInteractions.keyDownOn(getInput(), 33, modifiers);
+          MockInteractions.keyDownOn(overlay(), 33, modifiers);
         }
 
         function ctrl() {
-          MockInteractions.keyDownOn(getInput(), 17);
+          MockInteractions.keyDownOn(overlay(), 17);
         }
 
         function enter() {
-          MockInteractions.pressEnter(getInput());
+          MockInteractions.pressEnter(overlay());
         }
 
         function space() {
-          MockInteractions.pressSpace(getInput());
+          MockInteractions.pressSpace(overlay());
         }
 
-        function getInput() {
+        function input() {
           return datepicker.$.input;
+        }
+
+        function overlay() {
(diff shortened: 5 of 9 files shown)


Saulis:
I think we need to add something also to navigate months or years quickly, page down/up maybe?

tehapo:
Looks really nice! But noticed that you might get several month animations going on if you are fast with pressing down/up arrow keys near a month boundary. Then it would appear to be a choppy animation instead of the smooth transition when only pressing once to go to next/prev month.

tomivirkki:
Would make sense to have the keyboard functionality related to overlay contents inside the overlay itself and enter/esc bindings in the vaadin-date-picker level. That would separate concerns to correct modules.

Saulis:
@tehapo I'll see what I can do
@tomivirkki I see what you mean, but not sure if it's feasible to move the handlers to overlay. For example 'down' and 'up' handlers need to open the overlay and not sure if that can be done from inside? Another problem is that overlay (after a really quick try) seems to need focus in order for the keydown events to be registered there.

---

Review status: 0 of 6 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/78#-:-KBm-QowY_r0NNmalLbM:1978778329)*

Saulis:
@tehapo added a check that will make it scroll only when month changes. 

As a general note, I decided to use Shift inside of Control (as stated in the WAI-ARIA specs) for now. Reason being, that at least on my comp hitting Control and arrow keys move windows around and I'm not so sure at this point how to deal with that. We should maybe look into it in the ARIA support task.

---

Review status: 0 of 6 files reviewed at latest revision, all discussions resolved.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/78#-:-KBm25KIFGny9vZL8SVp:-1054326459)*

jounik:
Quickly tried it out. Two notices:

- As discussed with others, it would be nicer if the list is only scrolled when necessary, so that we don’t always scroll the selected month to the top of the dropdown
- When blurring the field, the dropdown should close. Now you can get two popups open at the same time when tabbing out of one date picker. This is not completely in the scope of this PR, though

tehapo:
Reviewed 2 of 6 files at r1, 4 of 4 files at r2.
Review status: all files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[vaadin-date-picker.html, line 342 \[r2\]](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/78#-KBqZAETK7IB8IwpFO5J:-KBqZAEUmyD9Cm4q8etF:2098383244) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/9824d47e54c174ce4a78aabdbdcc262a9c0b5549/vaadin-date-picker.html#L342)):*
Why not just do a direct assignment instead of using the `set` method?

---

*[vaadin-date-picker.html, line 348 \[r2\]](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/78#-KBqZFePEjSaKQ1CyVAy:-KBqZFePEjSaKQ1CyVAz:2098383244) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/9824d47e54c174ce4a78aabdbdcc262a9c0b5549/vaadin-date-picker.html#L348)):*
Why not just do a direct assignment instead of using the `set` method?

---

*[vaadin-date-picker.html, line 350 \[r2\]](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/78#-KBqZGFB3B-Dyb5foJgA:-KBqZGFCVcA9803GyI4A:2098383244) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/9824d47e54c174ce4a78aabdbdcc262a9c0b5549/vaadin-date-picker.html#L350)):*
Why not just do a direct assignment instead of using the `set` method?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/78)*

tehapo:
Related to the comment by @jounik – now that I have tried it, I actually like that the focused month is always scrolled to top.

Saulis:
I've changed the scrolling now to scroll only when focus goes hidden. About the blurring, I would consider doing that separately since it concerns blurring in general, not just with tab.

---

Review status: 3 of 6 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[vaadin-date-picker.html, line 342 \[r2\]](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/78#-KBqZAETK7IB8IwpFO5J:-KBr5ye_eOi_hbpONlo4:-499146009) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/9824d47e54c174ce4a78aabdbdcc262a9c0b5549/vaadin-date-picker.html#L342)):*
Done.

---

*[vaadin-date-picker.html, line 348 \[r2\]](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/78#-KBqZFePEjSaKQ1CyVAy:-KBr5zAcz1FwppmOJAQP:-499146009) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/9824d47e54c174ce4a78aabdbdcc262a9c0b5549/vaadin-date-picker.html#L348)):*
Done.

---

*[vaadin-date-picker.html, line 350 \[r2\]](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/78#-KBqZGFB3B-Dyb5foJgA:-KBr5zXpnQZbLfE9oeI4:-499146009) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/9824d47e54c174ce4a78aabdbdcc262a9c0b5549/vaadin-date-picker.html#L350)):*
Done.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/78#-:-KBr6sNog3KjEk_UEYiS:755188288)*

tomivirkki:
Some focus issue still remains. If I click "today" while navigating with keys, the component loses focus and keyboard navigation doesn't work anymore.

---

Review status: 3 of 6 files reviewed at latest revision, 4 unresolved discussions, some commit checks broke.

---

*[vaadin-date-picker-overlay.html, line 339 \[r3\]](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/78#-KBrmUSCW4qVOicsyq5F:-KBrmUSDzVigs_8_JM6C:1068993736) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/2ccdbcd2d9315d73b2c66d2633b2fef944088143/vaadin-date-picker-overlay.html#L339)):*
Leaving out Math.floor would always scroll so that the focused month is at the bottom (when going downwards). This would make it behave similarly as when going upwards.

This is just an alternative, not saying that it should be changed to that if the current has better feel to it.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/78#-:-KBrnef6TmZzVCbmddPn:-932820669)*

tehapo:
The focus issue @tomivirkki mentioned is still there. And a rebase is needed. Otherwise LGTM.

---

Reviewed 1 of 3 files at r3, 4 of 4 files at r4.
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/78#-:-KBw7vAS5dhdOY3xCcUU:-1106641884)*

Saulis:
OK, I've rebased, moved the keydown handlers to overlay, forced the focus on overlay also after clicking on "Today". Oh and also added styling properties for the colors - not sure if the default values are good or not.

Also, I'd like to hear your opinions on if there should always be a focused date visible, and not after the first keypress.

---

Review status: 2 of 9 files reviewed at latest revision, 1 unresolved discussion.

---

*[vaadin-date-picker-overlay.html, line 339 \[r3\]](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/78#-KBrmUSCW4qVOicsyq5F:-KBwKQdcj36n2TnP8JDT:856119268) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/2ccdbcd2d9315d73b2c66d2633b2fef944088143/vaadin-date-picker-overlay.html#L339)):*
Well that's true. I think I'll remove it.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/78#-:-KBwKi2yZGPsD2x0mXss:1354763441)*

tehapo:
Reviewed 7 of 7 files at r5.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[vaadin-date-picker.html, line 415 \[r5\]](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/78#-KByAO9q4KPxeh1mL69o:-KByAO9rOgxNOzr1bybm:571037964) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/91ec7dfbef27ad429ac2374814f38e66dc7e5bd3/vaadin-date-picker.html#L415)):*
The `case`s in this `switch` call `_open()` which was actually already changed to `open()` (in a commit that you now rebased on).

Also, as the body of both are identical, I would combine these. It would result in:

```
case 'down':
case 'up':
  // prevent scrolling the page with arrows
  e.preventDefault();

  if (!this._opened) {
    this.open();
  }
  break;
```

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/78)*

Saulis:
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[vaadin-date-picker.html, line 415 \[r5\]](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/78#-KByAO9q4KPxeh1mL69o:-KByCB5GoBqSZ4P4e1Qw:-1717687390) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/91ec7dfbef27ad429ac2374814f38e66dc7e5bd3/vaadin-date-picker.html#L415)):*
Yeah, I rebased but accidentally left these changes out of the commit :8ball: 

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/78)*

tehapo:
:lgtm: 

---

Reviewed 2 of 2 files at r6.
Review status: 7 of 9 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/78#-:-KC-Z3g1ux7HOKhVUR1E:980077423)*

tehapo:
Reviewed 2 of 2 files at r7.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/78)*

PR #81, opened 2016-03-03
Tweak vaadin-infinite-scroller update thresholds

tehapo:
This small change will move the update threshold of the scroller to make sure we always have at least `buffer size - 1` items visible. Previously we had `buffer size / 2` as the minimum of visible items and it might sometimes lead to empty area being visible when the `vaadin-date-picker` was used inside a very high viewport.

[Review on Reviewable](https://reviewable.io/reviews/vaadin/vaadin-date-picker/81)

### DIFF
--- a/vaadin-date-picker-overlay.html
+++ b/vaadin-date-picker-overlay.html
@@ -255,7 +255,7 @@
           <vaadin-month-calendar locale="[[locale]]" month="[[_dateAfterXMonths(index)]]" selected-date="{{selectedDate}}"></vaadin-month-calendar>
         </template>
       </vaadin-infinite-scroller>
-      <vaadin-infinite-scroller id="yearScroller" on-tap="_onYearTap" on-scroll="_onYearScroll" on-touchstart="_onYearScroll" item-height="150" buffer-size="50">
+      <vaadin-infinite-scroller id="yearScroller" on-tap="_onYearTap" on-scroll="_onYearScroll" on-touchstart="_onYearScroll" item-height="150" buffer-size="10">
         <template>
           <div>[[_yearAfterXYears(index)]]</div>
         </template>
--- a/vaadin-infinite-scroller.html
+++ b/vaadin-infinite-scroller.html
@@ -136,8 +136,9 @@ This program is available under Apache License Version 2.0, available at https:/
       }
 
       // Check if we scrolled enough to translate the buffer positions.
-      var upperThresholdReached = scrollTop > this._buffers[1].translateY + this._bufferHeight / 2;
-      var lowerThresholdReached = scrollTop < this._buffers[0].translateY + this._bufferHeight / 2;
+      var bufferOffset = this.$$('.buffer').offsetTop;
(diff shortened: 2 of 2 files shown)


tomivirkki:
This improves the situation on month scroller but creates an inverted problem with year scroller
https://www.dropbox.com/s/1688duiarbhuaqo/Screenshot%202016-03-03%2009.22.52.png?dl=0

This is caused by the 50% addition to year scroller buffer top. In theory this could be avoided by adding the effective buffer top value to the calculation but it might get messy.

It's not as big a problem as the one this resolves so I'd give this a LGTM as it is.

---







*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/81#-:-KBvN61gcEln5MeVw3Qt:1681323321)*

tehapo:
Just pushed a new version to fix the situation of the screenshot by taking the top offset into account.

Also reduced the buffer size of the year scroller to 10 (down from 50).

---

Review status: 0 of 2 files reviewed at latest revision, all discussions resolved.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/81#-:-KBvknVXIOQHhbvOPmCd:-1071470707)*

tomivirkki:
*[vaadin-infinite-scroller.html, line 141 \[r2\]](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/81#-KBvrWvitBcQf6NjPUnL:-KBvrWvitBcQf6NjPUnM:516158848) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/9dc4aad72dcfcb2643f8a49c145f3886453f8a66/vaadin-infinite-scroller.html#L141)):*
Should we just compute this on every loop?
The height of the component can vary dynamically so bufferoffset should change as well

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/81)*

tehapo:
Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion.

---

*[vaadin-infinite-scroller.html, line 141 \[r2\]](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/81#-KBvrWvitBcQf6NjPUnL:-KBvzihZgS8u9gDo8aqQ:1737116692) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/9dc4aad72dcfcb2643f8a49c145f3886453f8a66/vaadin-infinite-scroller.html#L141)):*
You're right. I was just trying to avoid any unnecessary work inside the scroll handler. Fixed.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/81)*

tomivirkki:
Reviewed 1 of 2 files at r2, 1 of 1 files at r3.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/81)*

PR #85, opened 2016-03-04
Prevent resetting initial position with a new date when reopening

Saulis:
Related with #76 

Assigning a new Date() value has a different timestamp everytime, so it will
always fire initialPositionChanged on reopen if we don't reuse it.

In English: the problem currently is that when you do keyboard navigation and close + reopen the overlay, it's scrolled back to Today if no initialPosition is set but the keyboard focus still stays where it was before closing.

[Review on Reviewable](https://reviewable.io/reviews/vaadin/vaadin-date-picker/85)

### DIFF
--- a/test/basic.html
+++ b/test/basic.html
@@ -94,6 +94,25 @@
         datepicker.open();
       });
 
+      it('should remember the original initial position on reopen', function(done) {
+        datepicker.initialPosition = null;
+
+        datepicker.open();
+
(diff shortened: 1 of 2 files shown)


tomivirkki:
Reviewed 2 of 2 files at r1.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/85)*

PR #86, opened 2016-03-04
Use Date.parse

manolo:
Use native Date parse instead of custom one.

[Review on Reviewable](https://reviewable.io/reviews/vaadin/vaadin-date-picker/86)

### DIFF
--- a/vaadin-date-picker.html
+++ b/vaadin-date-picker.html
@@ -235,17 +235,11 @@ See paper-input-container documentation for styling the included input fields
       },
 
       _parseDate: function(str) {
-        // Parsing with RegExp to ensure correct format
-        var parts = /^(\d{4}|[-+]\d{6})-(\d{2})-(\d{2})$/.exec(str);
-        if (!parts) {
-          return;
+        // We only accept ISO format without time info: (+-YY)YYYY-MM-DD
+        if (/^([-+]\d\d)?\d{4}(-\d\d){2}$/.test(str)) {
+          // Date.parse returns time in UTC, so we add the local offset
+          return new Date(Date.parse(str) + new Date().getTimezoneOffset() * 60000);
         }
-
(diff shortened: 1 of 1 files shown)


platosha:
Personally, I'd prefer to go with the current custom parser instead of the `Date.parse` solution. Its verbose code isn't necessarily bad: it's also easier to understand and support, this potentially leads to less errors.

---

Reviewed 1 of 1 files at r1.

---

*[vaadin-date-picker.html, line 241 \[r1\]](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/86#-KC0hKabfyWhR788RiAy:-KC0hKabfyWhR788RiAz:291826127) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/8952767b3775ecacd09d9a8e38a5619dc333d30a/vaadin-date-picker.html#L241)):*
This is erroneous. Add a timezone offset of today to a given date, which might have a different timezone offset (e.g., daylight saving time, timezone migrations and so on). Examples:

Correct result (today is the same timezone offset with the given date):
```javascript
new Date(Date.parse('2010-01-01') + new Date().getTimezoneOffset() * 60000)
// Fri Jan 01 2010 00:00:00 GMT+0200 (EET)
```

Wrong result (today is winter, the given date is on summer):
```javascript
new Date(Date.parse('2010-06-01') + new Date().getTimezoneOffset() * 60000)
// Tue Jun 01 2010 01:00:00 GMT+0300 (EEST)
```

You've got to subtract the timezone offset of that date from the date.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/86#-:-KC0g5UgGKvZkqJyTFNr:-1860038693)*

manolo:
There is an option to make it work with native parsing, but you are right code would be lesser and more complex to follow. Abandoning the PR.

PR #132, opened 2016-03-21
Make sure the scroller has visible content once scrolling stops

tomivirkki:
If you scroll (the month scroller) fast enough on a slower mobile device, you'll probably notice the scroller content appearing empty at some point. This is because the buffers are out of the view due to really fast scrolling.

Normally the scroller recovers from this once the momentum slows down but if you'll for example stop the momentum with a finger, you might end up with an empty scroller.

This PR makes sure that the scroller has visible content every time scrolling stops.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-date-picker/132)

### DIFF
--- a/vaadin-infinite-scroller.html
+++ b/vaadin-infinite-scroller.html
@@ -150,6 +150,13 @@ This program is available under Apache License Version 2.0, available at https:/
         this._mayHaveMomentum = true;
       }
       this._preventScrollEvent = false;
+
+      this.debounce('scrollFinish', function() {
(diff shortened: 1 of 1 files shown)


tehapo:
:lgtm:

---

Reviewed 1 of 1 files at r1.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/132#-:-KDNQD-wImmLEP4Bhkim:1417088721)*

PR #162, opened 2016-04-11
Align year numbers better with months

jounik:
Align the year numbers with the arrow indicator together with the current month in the month scroller. E.g. for January, the arrow points roughly to the center of the year number text.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-date-picker/162)

### DIFF
--- a/vaadin-date-picker-overlay.html
+++ b/vaadin-date-picker-overlay.html
@@ -117,6 +117,7 @@
         transform: translateX(100%);
         -webkit-tap-highlight-color: transparent;
         @apply(--paper-font-menu);
+        line-height: 8px;
         @apply(--vaadin-date-picker-yearscroller);
       }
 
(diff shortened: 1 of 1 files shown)


tomivirkki:
We've usually used the following for adjusting the very same thing
https://github.com/vaadin/vaadin-date-picker/blob/master/vaadin-date-picker-overlay.html#L139

Wondering why it was off :/

jounik:
Oh right, didn’t notice/remember. So I should update to use that then, right?

tomivirkki:
I guess it would make more sense yeah.

tomivirkki:
Reviewed 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/162)*

PR #214, opened 2016-07-27
Add autoValidate property declaration

### LINKED ISSUE #213: `autoValidate` property is bound but not declared
Binding for undeclared properties works, but the problem is that the property is not documented and will not be reflected in generated APIs.

There might be other non-declared properties as well. We should declare them all and add documentation.

platosha:
Fixes #213

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-date-picker/214)

### DIFF
--- a/vaadin-date-picker.html
+++ b/vaadin-date-picker.html
@@ -227,6 +227,14 @@ If you want to replace the default input field with a custom implementation, you
 
       properties: {
         /**
+         * Set to true to auto-validate the input value.
+         */
+        autoValidate: {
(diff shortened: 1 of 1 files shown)


manolo:
:lgtm:

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/214#-:-KNfuAgY0fXcHlBXUH-z:1417088721)*

PR #244, opened 2016-09-29
Add property and attribute to hide floating label

samcarecho:
vaadin-date-picker does not allow to hide the floating label.

Solution: 
Add the property "noLabelFloat" and the attribute "no-label-float".

After implementing the solution, adding the attribute "no-label-float" to <vaadin-date-picker> will hide the floating label.

I also changed the indent of part of the code to make it easier to visualize the attributes.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-date-picker/244)

### DIFF
--- a/vaadin-date-picker.html
+++ b/vaadin-date-picker.html
@@ -199,7 +199,7 @@ If you want to replace the default input field with a custom implementation, you
     </paper-input-container>
 
     <iron-dropdown id="dropdown"
-                   fullscreen$=[[_fullscreen]]
+                   fullscreen$="[[_fullscreen]]"
                    allow-outside-scroll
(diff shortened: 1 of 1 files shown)


tomivirkki:
Hi @samcarecho 
Thanks for the PR, the implementation looks good.

However, we provide `<vaadin-date-picker-light>` for this very purpose. Basically instead of duplicating all `<paper-input>` properties to `<vaadin-date-picker>` public API, we provide a 'light' version of the component in which you can place a custom `<paper-input>`.

More information at https://vaadin.com/docs/-/part/elements/vaadin-date-picker/vaadin-date-picker-customization.html

PR #297, opened 2016-11-17
Ux fixes v1.2

### LINKED ISSUE #256: Keyboard navigation is confusing with keyboard input
The following issues with the keyboard navigation make it complex to understand, and therefore, to explain it for the screen reader users.
### Scenario 1
1. Click an empty date picker. The date picker gets focused, and the overlay opens.
2. Press <kbd>←</kbd> or <kbd>→</kbd> keys.

Or, alternatively:
1. Focus an empty date picker with a <kbd>Tab</kbd> key. The overlay stays closed.
2. Press <kbd>↓</kbd> or <kbd>↑</kbd> key to open the overlay.
3. Press <kbd>←</kbd> or <kbd>→</kbd> keys.
#### Issue 1.1:

Expected: the focused date in the overlay is changed to the previous or the next date.

Actual: the focused date will not change until the <kbd>↓</kbd> or <kbd>↑</kbd> key is pressed.

Apparently, when the input field is in focus and the overlay is opened, there are could be two modes for the cursor keys:
1. The cursor mode (default): the keys move the cursor in the input field.
2. The calendar mode: they move the focused date in the overlay.

Both modes have reasons to exist, perhaps. What is a bit confusing here is that the first mode is used on empty input fields, where it doesn’t make much sense to move the input cursor. 
#### Issue 1.2:

It is harder to select yesterday or tomorrow with arrow keys. In practice, you have to:
1. Press <kbd>↓</kbd> to activate calendar mode. This also jumps a week in future.
2. Press <kbd>↑</kbd> to jump back to the initial date.
3. Press <kbd>←</kbd> or <kbd>→</kbd> to finally jump to yesterday or tomorrow.
#### Issue 1.3:

The cursor keys mode is not indicated in the overlay. The focused date in the calendar has the same appearance in both cases: when it is controlled by the <kbd>←</kbd> or <kbd>→</kbd> keys and when it is not.
### Scenario 2
1. Open the overlay on an empty date picker.
2. Press <kbd>↓</kbd> key.
3. Press <kbd>Esc</kbd> key multiple times until the overlay gets closed.

Or:
1. Open the overlay on date picker with some pre-selected date.
2. Press <kbd>↓</kbd> key.
3. Type “asdf” or some other non-date text.
4. Press <kbd>↓</kbd> key once again.
5. Press <kbd>Esc</kbd> key multiple times until the overlay gets closed.
#### Issue

Expected: the overlay closes after one or two <kbd>Esc</kbd> key hits.

Actual: It is required to press <kbd>Esc</kbd> three times to close the overlay.
### Solution ideas
1. Always use calendar cursor mode when the overlay is opened. For screen reader users, announce how to open the overlay on the input field, announce cursor keys behavior when the overlay opens.
2. Introduce additional focus-like indication on the overlay, which will tell the user if the arrow keys control the overlay or the input field cursor. For screen reader users, requires additional announce when the mode changes.

More UX test results in https://github.com/vaadin/components-team-tasks/issues/212

### LINKED ISSUE #275: Clicking the overlay steals focus from the input field
After that keyboard input/navigation doesn't work

### LINKED ISSUE #291: Implement UX changes for 1.2 final
- Esc should work as 'cancel'
- Today/Cancel buttons should be focusable with tab
- Mousedown should focus a date
- Focus should be indicated with proper theming
- Should not focus/select the input text while navigating with arrows
- Right/left arrows should also move focus to overlay iff input field is empty

### LINKED ISSUE #296: Date in the input is selected as text after choosing the date
### Description
Date in the input is selected as text after choosing the date 

### Expected outcome
Text is not selected, it's focused and the cursor is at the end

### Actual outcome
recorded

### Browsers Affected
- [ ] Chromium on Linux

tomivirkki:
Fixes #291, Fixes #296, Fixes #256, Fixes #275

- Esc works as cancel
- Left/right arrows move focus to overlay from an empty input field
- Today/Cancel buttons can be tabbed to
- Date cell has :active style
- Clearer separation between input/overlay focus

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-date-picker/297)

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -68,8 +68,8 @@ function testSauce(browsers, done) {
 
 gulp.task('test:desktop', function(done) {
   testSauce([
-    'Windows 10/chrome@48',
-    'Windows 10/firefox@44',
+    'Windows 10/chrome@50',
+    'Windows 10/firefox@47',
     'Windows 10/microsoftedge@13',
     'Windows 10/internet explorer@11',
     'OS X 10.11/safari@9.0'], done);
@@ -116,8 +116,8 @@ gulp.task('test:desktop:shadow', function(done) {
   args.dom = 'shadow';
 
   testSauce([
-    'Windows 10/chrome@48',
-    'Windows 10/firefox@44',
+    'Windows 10/chrome@50',
+    'Windows 10/firefox@47',
     'Windows 10/microsoftedge@13',
     //'Windows 10/internet explorer@11', // shadow polyfill seems to have issues in IE11.
     'OS X 10.11/safari@9.0'], done);
--- a/test/basic.html
+++ b/test/basic.html
@@ -259,14 +259,15 @@
         });
 
         it('should realign on window scroll', function(done) {
-          var spy = sinon.spy(datepicker, '_updateAlignmentAndPosition');
+          sinon.stub(datepicker, '_updateAlignmentAndPosition', function() {
+            if (!done._called) {
+              done._called = true;
+              done();
+            }
+          });
           Polymer.Base.fire('scroll', {}, {
             node: window
           });
-          datepicker.async(function() {
-            expect(spy.called).to.be.true;
-            done();
-          }, 100);
         });
 
         it('should realign on container scroll', function(done) {
@@ -275,16 +276,15 @@
           var container = fixture('datepicker-wrapped');
           var datepickerWrapped = container.querySelector('vaadin-date-picker');
 
-          datepickerWrapped.addEventListener('iron-overlay-opened', function() {
-            var spy = sinon.spy(datepickerWrapped, '_updateAlignmentAndPosition');
+          open(datepickerWrapped, function() {
+            sinon.stub(datepickerWrapped, '_updateAlignmentAndPosition', function() {
+              if (!done._called) {
+                done._called = true;
+                done();
+              }
+            });
             container.scrollTop += 100;
-
-            datepickerWrapped.async(function() {
-              expect(spy.called).to.be.true;
-              done();
-            }, 100);
           });
-          datepickerWrapped.open();
         });
 
         it('should not realign on year/month scroll', function(done) {
--- a/test/basic.html
+++ b/test/basic.html
@@ -384,12 +384,13 @@
           datepicker.set('i18n.today', 'Tänään');
           datepicker.set('i18n.cancel', 'Peruuta');
 
-          datepicker.$.overlay.$.scroller.addEventListener('animationend', function() {
-            datepicker.debounce('animations-finished', function() {
-              done();
-            }, 1);
+          open(datepicker, function() {
+            Polymer.RenderStatus.afterNextRender(datepicker.$.overlay.$.scroller, function() {
+              Polymer.Base.async(function() {
+                done();
+              }, 1);
+            });
           });
-          datepicker.open();
 
         });
 
--- a/test/basic.html
+++ b/test/basic.html
@@ -50,54 +50,46 @@
         expect(spy.calledOnce).to.be.true;
       });
 
+      it('should open with open call', function(done) {
+        open(datepicker, function() {
+          done();
+        });
+      });
+
       it('should notify opened changed on open', function(done) {
-        datepicker.addEventListener('opened-changed', function() {
+        open(datepicker, function() {
           expect(datepicker.opened).to.be.true;
           done();
         });
-        datepicker.open();
       });
 
       it('should notify opened changed on close', function(done) {
-        datepicker.addEventListener('iron-overlay-opened', function() {
-          datepicker.addEventListener('opened-changed', function() {
+        open(datepicker, function() {
+          listenForEvent(datepicker, 'opened-changed', function() {
             expect(datepicker.opened).to.be.false;
             done();
           });
           datepicker.close();
         });
-        datepicker.open();
-      });
-
-      it('should open with open call', function(done) {
-        datepicker.addEventListener('iron-overlay-opened', function() {
-          done();
-        });
-        datepicker.open();
       });
 
       it('should close with close call', function(done) {
-        datepicker.addEventListener('iron-overlay-opened', function() {
-          datepicker.close();
+        open(datepicker, function() {
+          close(datepicker, function() {
+            expect(datepicker.opened).to.be.false;
+            done();
+          });
         });
-        datepicker.addEventListener('iron-overlay-closed', function() {
-          done();
-        });
-        datepicker.open();
       });
 
       it('should open on input tap', function(done) {
-        datepicker.addEventListener('iron-overlay-opened', function() {
-          done();
-        });
+        listenForEvent(datepicker, 'iron-overlay-opened', done);
         tap(datepicker.$.input);
       });
 
       it('should open on label tap', function(done) {
-        datepicker.addEventListener('iron-overlay-opened', function() {
-          done();
-        });
-        tap(datepicker.$.label);
+        listenForEvent(datepicker, 'iron-overlay-opened', done);
+        tap(datepicker.$.input);
       });
 
       it('should pass the placeholder attribute to the input tag', function() {
@@ -117,11 +109,10 @@
 
       it('should scroll to today by default', function(done) {
         var spy = sinon.spy(datepicker.$.overlay, 'scrollToDate');
-        datepicker.addEventListener('iron-overlay-opened', function() {
+        open(datepicker, function() {
           expect(monthsEqual(spy.firstCall.args[0], new Date())).to.be.true;
           done();
         });
-        datepicker.open();
       });
 
       it('should scroll to initial position', function(done) {
@@ -129,11 +120,10 @@
         var initialPositionDate = new Date(2016, 0, 1);
 
         var spy = sinon.spy(datepicker.$.overlay, 'scrollToDate');
-        datepicker.addEventListener('iron-overlay-opened', function() {
+        open(datepicker, function() {
           expect(spy.firstCall.args[0]).to.eql(initialPositionDate);
           done();
         });
-        datepicker.open();
       });
 
       it('should remember the original initial position on reopen', function(done) {
@@ -156,24 +146,21 @@
       it('should scroll to selected value by default', function(done) {
         var spy = sinon.spy(datepicker.$.overlay, 'scrollToDate');
         datepicker.value = '2000-02-01';
-        datepicker.addEventListener('iron-overlay-opened', function() {
+        open(datepicker, function() {
           expect(monthsEqual(spy.firstCall.args[0], new Date(2000, 1, 1))).to.be.true;
           done();
         });
-        datepicker.open();
       });
 
       it('should close on overlay date tap', function(done) {
-        datepicker.addEventListener('iron-overlay-closed', function() {
-          done();
-        });
-        datepicker.addEventListener('iron-overlay-opened', function() {
+        listenForEvent(datepicker, 'iron-overlay-closed', done);
+
+        open(datepicker, function() {
           Polymer.Base.fire('date-tap', {}, {
             bubbles: true,
             node: datepicker.$.overlay
           });
         });
-        datepicker.open();
       });
 
       it('should not have label defined by default', function() {
@@ -215,7 +202,12 @@
 
       it('should show clear icon', function(done) {
         datepicker.value = '2000-02-01';
-        datepicker.addEventListener('iron-overlay-opened', function() {
+        Polymer.Base.fire('date-tap', {}, {
+          bubbles: true,
+          node: datepicker.$.overlay
+        });
+
+        open(datepicker, function() {
           if (isFullscreen(datepicker)) {
             expect(datepicker.$.overlay.$.clear.clientHeight).not.to.equal(0);
           } else {
@@ -223,35 +215,31 @@
           }
           done();
         });
-        datepicker.open();
       });
 
       it('should open by tapping the calendar icon', function(done) {
+        listenForEvent(datepicker, 'iron-overlay-opened', done);
         tap(datepicker.$.calendar);
-        datepicker.addEventListener('iron-overlay-opened', function() {
-          done();
-        });
       });
 
       it('should scroll to a date on open', function(done) {
         // We must scroll to a date on every open because at least IE11 seems to reset
         // scrollTop while the dropdown is closed. This will result in all kind of problems.
         var spy = sinon.spy(datepicker.$.overlay, 'scrollToDate');
-        var openTimes = 0;
-        datepicker.addEventListener('iron-overlay-opened', function() {
+
+        open(datepicker, function() {
           expect(spy.called).to.be.true;
+
           spy.reset();
-          if (openTimes === 2) {
-            done();
-          } else if (openTimes < 2) {
-            datepicker.close();
-          }
-          openTimes++;
+          close(datepicker, function() {
+
+            open(datepicker, function() {
+              expect(spy.called).to.be.true;
+              done();
+            });
+          });
+          datepicker.close();
         });
-        datepicker.addEventListener('iron-overlay-closed', function() {
-          datepicker.open();
-        });
-        datepicker.open();
       });
 
       it('should not change datepicker width', function() {
@@ -267,12 +255,7 @@
       describe('window scroll realign', function() {
 
         beforeEach(function(done) {
-          datepicker.addEventListener('iron-overlay-opened', function() {
-            datepicker.async(done, 100);
-          });
-          datepicker.async(function() {
-            datepicker.open();
-          }, 100);
+          open(datepicker, done);
         });
 
         it('should realign on window scroll', function(done) {
--- a/test/common.js
+++ b/test/common.js
@@ -25,6 +25,24 @@ function getDefaultI18n() {
   };
 }
 
+function listenForEvent(elem, type, callback) {
+  var listener = function() {
+    elem.removeEventListener(type, listener);
+    callback();
+  };
+  elem.addEventListener(type, listener);
+}
+
+function open(datepicker, callback) {
+  listenForEvent(datepicker, 'iron-overlay-opened', callback);
+  datepicker.open();
+}
+
+function close(datepicker, callback) {
+  listenForEvent(datepicker, 'iron-overlay-closed', callback);
+  datepicker.close();
+}
+
 function tap(element) {
   Polymer.Base.fire('tap', {}, {
     bubbles: true,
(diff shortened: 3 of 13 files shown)


manolo:
differences with the prototype:
- moving arrows when the overlay is not focused should not move the calendar
- when focusing the input does not select all the text in the input
- when clicking on one date the previous focused continues background-ed only one date should be active

---

Review status: 0 of 13 files reviewed at latest revision, 1 unresolved discussion, some commit checks broke.

---

*[vaadin-date-picker-behavior.html, line 367 at r3](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/297#-KWmphNtd8ALIV2TLG5A:-KWmphNtd8ALIV2TLG5B:b-c39m6b) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/25d0c8647114f6056a15b79c8bfab5ec8a2093a4/vaadin-date-picker-behavior.html#L367)):*
> ```HTML
>       if (!this._ignoreFocusedDateChange && !this._noInput && this.opened) {
>         this._inputValue = focusedDate ? formatDate(focusedDate) : '';
>         this._input().setSelectionRange(0, this._inputValue.length);
> ```

The selection range should be applied when focusing the input again, so as it was so in the prototype, and its the default behaviour of inputs.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/297#-:-KWmr9traCEw2YPfAQx9:b-aqx07a)*

manolo:
More issues when focusing the overlay the focused date should be in the viewport

---

Review status: 0 of 13 files reviewed at latest revision, 1 unresolved discussion, some commit checks broke.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/297#-:-KWms8yc24YwBXKXN4bW:b-cz2pzf)*

manolo:
I don't know if this should be fixed, but running the example in desktop with a short height , the mobile version of the overlay is displayed. When looping focus the input seems not be selected, or seems to have style problems

---

Review status: 0 of 13 files reviewed at latest revision, 1 unresolved discussion, some commit checks broke.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/297#-:-KWmsnUHMh7xb5OyF5cZ:b47brm7)*

tomivirkki:
> moving arrows when the overlay is not focused should not move the calendar
You mean when the input field is empty and you hit right/left? We agreed that this is how it should work.

> when focusing the input does not select all the text in the input
Works for me. Can you be more specific.

> when clicking on one date the previous focused continues background-ed only one date should be active
Active != focused. We shouldn't move the focus on keydown.

> More issues when focusing the overlay the focused date should be in the viewport
Please rephrase. I'm not following.

> I don't know if this should be fixed, but running the example in desktop with a short height , the mobile version of the overlay is displayed. When looping focus the input seems not be selected, or seems to have style problems
Doesn't relate to this PR. Our mobile detection could be improved later if it causes issues.

---

Review status: 0 of 13 files reviewed at latest revision, 1 unresolved discussion, some commit checks broke.

---

*[vaadin-date-picker-behavior.html, line 367 at r3](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/297#-KWmphNtd8ALIV2TLG5A:-KWn5_fJYSODSyFgnEjr:bogn5hr) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/25d0c8647114f6056a15b79c8bfab5ec8a2093a4/vaadin-date-picker-behavior.html#L367)):*
<details><summary><i>Previously, manolo (Manuel Carrasco Moñino) wrote…</i></summary>
> The selection range should be applied when focusing the input again, so as it was so in the prototype, and its the default behaviour of inputs.
</details>
Works for me. Can you be more specific.
This (removed) line is about setting the selection range on keyboard navigation.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/297#-:-KWn4bVR0DkDJTAlvqZJ:bzhd9r7)*

manolo:
Sure, it's difficult to explain with sentences. I guess a visual explanation should be better.

- Issue 1.
   - In the following image you can see that you can move the focused date although the ui focus is on the OK or Cancel buttons. You can prevent the calendar moving unless the scroller has the focus.
http://g.recordit.co/2aSrNczCsS.gif
   - In the prototype the issue does not happen.
http://g.recordit.co/NEntZVeZEE.gif

- Issue 2.
   - In the next screenshot, you can see that any browser input should select all the value when focused. But for the data picker input it does not happen if the focus comes from a focused element in the overlay
http://g.recordit.co/jQnIL5qjly.gif
   - In the prototype it works.
http://g.recordit.co/NfJml7AygN.gif

- Issue 3.
  -  Take a look to this, when I click on a day to select and maintain clicked for a while, there should only be a focused day, that must be the mouse-down-ded one. In the screenshot, the day to select does not have the appropriate background (focuses) which is gotten by the old day.
http://g.recordit.co/Ti46FWriX3.gif
  - Otherwise this is correct
http://g.recordit.co/aBiq9XsbHm.gif

- Isue 4
 - In this animation you can see that the focus is not visible when looping. I'm continuously  pushing the tab
http://g.recordit.co/DuFwUHB3iv.gif
 - This should be the appropriate behavior
http://g.recordit.co/6CnA3i6EJi.gif

 

---

Review status: 0 of 13 files reviewed at latest revision, 1 unresolved discussion, some commit checks broke.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/297#-:-KWnaxR7bojqLqhFDkoV:b-bio3jv)*

manolo:
Review status: 0 of 13 files reviewed at latest revision, 1 unresolved discussion, some commit checks broke.

---

*[vaadin-date-picker-behavior.html, line 367 at r3](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/297#-KWmphNtd8ALIV2TLG5A:-KWnvEvc3U2vakW7kbgt:b-ugin4x) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/25d0c8647114f6056a15b79c8bfab5ec8a2093a4/vaadin-date-picker-behavior.html#L367)):*
<details><summary><i>Previously, tomivirkki (Tomi Virkki) wrote…</i></summary>
> Works for me. Can you be more specific.
> This (removed) line is about setting the selection range on keyboard navigation.
</details>
Actually this must be removed to avoid moving the focus to the input. You cannot select a range of an input that is not focused.
What I meant is that you have to copy this in the programatically input focus code.
Take a look to the third picture in my comment above.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/297)*

tomivirkki:
Issue 1: Done.
Issue 2: Done.
Issue 3: focused != active, we're not supposed to move the focused date on mousedown but just highlight the active element.
Issue 4: Not related to this PR, we can improve mobile detection later if necessary. The prototype's behaviour isn't correct either: the input field in the header shouldn't be focusable, but again on mobile you won't be tabbing anyway so we don't need to worry about this now.

---

Review status: 0 of 13 files reviewed at latest revision, 1 unresolved discussion.

---

*[vaadin-date-picker-behavior.html, line 367 at r3](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/297#-KWmphNtd8ALIV2TLG5A:-KWqRfpoySZF7oIm-PaD:b-896fix) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/25d0c8647114f6056a15b79c8bfab5ec8a2093a4/vaadin-date-picker-behavior.html#L367)):*
<details><summary><i>Previously, manolo (Manuel Carrasco Moñino) wrote…</i></summary>
> Actually this must be removed to avoid moving the focus to the input. You cannot select a range of an input that is not focused.
> What I meant is that you have to copy this in the programatically input focus code.
> Take a look to the third picture in my comment above.
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/297#-:-KWqSXABPyveMyknBukQ:bgf32w1)*

manolo:
- issue 3: Please read https://www.flowdock.com/app/vaadin/elements-team/threads/otriNck02Il1M78Vg2JFeI3I2TH  from @jouni  `One more thing 😺 Would it be possible to give a visual feedback when I do mousedown on a day cell, so that the highlight/focus would change to that day cell? Or at least provide another visual style for a day cell that is being clicked?` and consequent commit in prototype : https://github.com/vaadin/vaadin-date-picker/commit/49085c15ce26e96f1aded5bc3ac234130e417a40

- issue 4: Right, I didn't say the the prototype worked, just wondering if it could be fixed somehow. I guess this should be fixed in this story since it full related (the focus issue), otherwise focus code could be disabled at all when the mobile version is shown in desktop.

---

Review status: 0 of 13 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/297#-:-KWqga5f3-85SnXJTk10:b-1te8h0)*

tomivirkki:
Issue 3: Discussed this personally with @jouni yesterday and we're not moving the focus but using active state instead, less intrusive
Issue 4: No need, focus is not considered in mobile mode. Correct approach would be to never snap to mobile view on desktop but it's not in scope of this PR

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/297#-:-KWqi0ySGmDtidnL-tDW:b-7mirof)*

manolo:
issue3: then was a lack of communication when changing reqs 🙁

---

Review status: 0 of 13 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/297#-:-KWqklNFU6uYGpizrbyO:brz7egk)*

jouni:
> issue3: then was a lack of communication when changing reqs 🙁

Sorry for any miscommunication (although I don’t think there was any). You did pretty much exactly what I requested, I wasn’t absolutely sure either how the behavior should be, and that’s why we prototype. Not sure if that aspect was communicated clearly enough, that we were still prototyping, and not looking for the final polished implementation.

tomivirkki:
Issue3: Well "Or at least provide another visual style for a day cell that is being clicked" kind of covers it also
Issue4: "otherwise focus code could be disabled at all when the mobile version is shown in desktop" Done (disabled tab key on fullscreen).

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/297#-:-KWqmE5lmMbV6gJWs5f5:b-lymbfj)*

manolo:
@jouni no problem.

There is another nit, when focusing the scroller, if the focused day is out of the viewport the overlay should move to show the focused day, otherwise the user does not have any feedback that the scroller is focused. Shouldn't this be fixed in this story ? 

---

Review status: 0 of 13 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/297#-:-KWqxCwAX0aA6FyESiCM:bmak6l5)*

tomivirkki:
Not necessarily expected behaviour, could be prototyped + UX tested. But anyway, this PR is about fixing the findings from the previously conducted UX tests. We'd like to get the beta out asap so let's not come up with more "might be nice to have" -stuff at this point.

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/297#-:-KWr-ydspd8z-cr0otGY:bhw45fg)*

manolo:
:lgtm:

---

Review status: 0 of 13 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/297#-:-KWr1G_GD8DUlbezANt6:bnfp4nl)*

manolo:
Reviewed 3 of 8 files at r1, 2 of 3 files at r2, 4 of 6 files at r3, 4 of 4 files at r4.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/297)*

tomivirkki:
> another nit

Got a second request for this feature, so included it also. Can you check if it works as you'd expect?

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/297#-:-KWr3CIHzEvuXDzeldrz:b-7wyoe1)*

manolo:
Cool. The fix works to me.

There is still a problem in IE11 that makes me tab twice to focus the scroller from the input when looping. It does not happen in chrome nor FF though.

---

Reviewed 1 of 1 files at r5.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/297#-:-KWr5ywYWAQymbVo9g1V:b-y3v46)*

tomivirkki:
Done. For some reason IE11 considers the svg element inside the calendar icon tabbable so needed to enforce the behaviour programmatically.

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/297#-:-KWrEKEarICdHydZxNtb:b-6bf95g)*

manolo:
Review status: 11 of 13 files reviewed at latest revision, 1 unresolved discussion.

---

*[vaadin-date-picker-behavior.html, line 648 at r6](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/297#-KWrKwZPm2tDjtWeR055:-KWrKwZPm2tDjtWeR056:b-4a0vqo) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/6970c882e793877286cd7da579cd20a3a5e302e5/vaadin-date-picker-behavior.html#L648)):*
> ```HTML
>             if (e.shiftKey) {
>               this.$.overlay.focusCancel();
>             } else {
> ```

I would do this only in the case of IE, so as someone reading the code understands

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/297)*

tomivirkki:
*[vaadin-date-picker-behavior.html, line 648 at r6](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/297#-KWrKwZPm2tDjtWeR055:-KWrLQBEVWku1-PKtiZ-:be7zbfr) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/6970c882e793877286cd7da579cd20a3a5e302e5/vaadin-date-picker-behavior.html#L648)):*
<details><summary><i>Previously, manolo (Manuel Carrasco Moñino) wrote…</i></summary>
> I would do this only in the case of IE, so as someone reading the code understands
</details>
Better to avoid browser-specific hacks. Further, this piece of code will be required for every browser anyway once we move the overlay to `<body>`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/297)*

manolo:
second :lgtm: , merge it if no one else has objections .

---

Reviewed 2 of 2 files at r6.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/297#-:-KWrR7JZRYiDBEQV_pgt:b-lj2zx9)*

platosha:
Review status: 12 of 13 files reviewed at latest revision, 2 unresolved discussions.

---

*[test/wai-aria.html, line 119 at r7](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/297#-KWrV91Tph2P3P67D1Me:-KWrV91Tph2P3P67D1Mf:bxv8aow) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/25d0c8647114f6056a15b79c8bfab5ec8a2093a4/test/wai-aria.html#L119)):*
> ```HTML
>           // tabindex makes the overlay focusable, breaks iOS VoiceOver mode:
>           // the VoiceOver cursor always jumps to the (focused) overlay
>           // when a user taps inside to hear what is under the finger.
> ```

Have you tried that it’s still usable with iOS VoiceOver before removing this?

---

*[test/wai-aria.html, line 444 at r7](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/297#-KWrUtvOBGOURpQneVda:-KWrUtvOBGOURpQneVdb:b-r9z42a) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/25d0c8647114f6056a15b79c8bfab5ec8a2093a4/test/wai-aria.html#L444)):*
> ```HTML
>         });
> 
>         it('should debounce announcements', function(done) {
> ```

How about if we keep this test to ensure that the event is only fired once? We can just rename it, now that there’s no debouncer anymore.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/297)*

tomivirkki:
Review status: 11 of 13 files reviewed at latest revision, 2 unresolved discussions.

---

*[test/wai-aria.html, line 119 at r7](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/297#-KWrV91Tph2P3P67D1Me:-KWrXU9h3eNrxBIqVFBA:b-coerk6) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/25d0c8647114f6056a15b79c8bfab5ec8a2093a4/test/wai-aria.html#L119)):*
<details><summary><i>Previously, platosha (Anton Platonov) wrote…</i></summary>
> Have you tried that it’s still usable with iOS VoiceOver before removing this?
</details>
Verified together just now..seems to work

---

*[test/wai-aria.html, line 444 at r7](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/297#-KWrUtvOBGOURpQneVda:-KWrYKs-mCl6lHNaHcmv:b-thgea0) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/25d0c8647114f6056a15b79c8bfab5ec8a2093a4/test/wai-aria.html#L444)):*
<details><summary><i>Previously, platosha (Anton Platonov) wrote…</i></summary>
> How about if we keep this test to ensure that the event is only fired once? We can just rename it, now that there’s no debouncer anymore.
</details>
Restored and renamed

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/297)*

platosha:
:lgtm:

---

Reviewed 1 of 1 files at r7, 1 of 1 files at r8.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/297#-:-KWrZ4v4Mvz4ZZudvxGq:bnfp4nl)*

platosha:
Reviewed 1 of 1 files at r9.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/297)*

PR #342, opened 2017-02-18
Remove override of padding-bottom on paper-input-container.

solidsystem:
**Problem**

When using date-picker element alongside normal paper-input elements, the padding-bottom of the container paper-input-container differ, causing the inputs to no align.

**Before fix**
vaadin-date-picker-padding-bug-problem

**After fix**
vaadin-date-picker-padding-bug-fixed

**Fix**
This commit removes the padding-bottom: 0 override of
paper-input-container, so that the date-picker aligns with the
height of  normal paper-input elements.

The override was introduced in commit
d888423def0ccc7869c801f3ef2315c14ff7b485. This change should probably be reviewed by it's author @tehapo before merging, since I guess there is a reason why it was added. Perhaps the original reason could be resolved differently.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-date-picker/342)

### DIFF
--- a/vaadin-date-picker.html
+++ b/vaadin-date-picker.html
@@ -85,7 +85,6 @@ If you want to replace the default input field with a custom implementation, you
         width: 420px;
       }
 
-
       #inputcontainer,
       #inputcontainer input {
         cursor: pointer;
--- a/vaadin-date-picker.html
+++ b/vaadin-date-picker.html
@@ -85,9 +85,6 @@ If you want to replace the default input field with a custom implementation, you
         width: 420px;
       }
 
-      #inputcontainer {
-        padding-bottom: 0;
-      }
 
       #inputcontainer,
       #inputcontainer input {


jouni:
Thanks for the PR!

I think the reason for removing the padding is purely visual, so that the overlay is aligned right under the input underline.

So, it would be great of you can make sure the alignment of the overlay is still correct after this change.

jouni:
Just to add – we shouldn’t have originally overridden any of the paper-input styles, so this fix is very welcome IMO.

andrewspy:
Any plan to fix the padding issue so that it can work well with paper elements?

manolo:
@jouni since we have moved the component to use themable-mixin, shouldn't we close this issue? and probably verify that vaadin-material-theme does not have the issue

jouni:
Well, if we want to keep 1.x backwards compatible and not break existing app layouts, we should not merge this.

But you’re right, we should try to not introduce similar issues in vaadin-material-theme for 2.0.

manolo:
Since the master branch has diverged so much, and paper-input container is not used anymore, this does not applies anymore. 

Thank you very much for your contribution though.

PR #352, opened 2017-04-13
Avoid passing null to bound elements. Fixes #351

### LINKED ISSUE #351: Dates do not get shown after polymer build
**Version**: 1.2.2
After a build, the actual dates are not visible to the end user:

screen shot 2017-04-11 at 2 52 31 pm

The app is built with polymer-cli@next with the following build options:
```
"builds": [{
    "name": "bundled",
    "addServiceWorker": true,
    "bundle": true,
    "js": {
      "compile": true,
      "minify": true
    },
    "css": {
      "minify": true
    },
    "html": {
      "minify": true
    }
  }]
```

manolo:
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-date-picker/352)

### DIFF
--- a/test/basic.html
+++ b/test/basic.html
@@ -351,23 +351,23 @@
         it('should not accept non-ISO formats', function() {
           datepicker.value = '03/02/01';
           expect(datepicker.value).to.equal('');
-          expect(datepicker._selectedDate).to.be.null;
+          expect(datepicker._selectedDate).to.equal('');
 
           datepicker.value = '2010/02/03';
           expect(datepicker.value).to.equal('');
-          expect(datepicker._selectedDate).to.be.null;
+          expect(datepicker._selectedDate).to.equal('');
 
           datepicker.value = '03/02/2010';
           expect(datepicker.value).to.equal('');
-          expect(datepicker._selectedDate).to.be.null;
+          expect(datepicker._selectedDate).to.equal('');
 
           datepicker.value = '3 Feb 2010';
           expect(datepicker.value).to.equal('');
-          expect(datepicker._selectedDate).to.be.null;
+          expect(datepicker._selectedDate).to.equal('');
 
           datepicker.value = 'Feb 3, 2010';
           expect(datepicker.value).to.equal('');
-          expect(datepicker._selectedDate).to.be.null;
+          expect(datepicker._selectedDate).to.equal('');
         });
 
         it('should output ISO format', function() {
--- a/vaadin-date-picker-behavior.html
+++ b/vaadin-date-picker-behavior.html
@@ -262,7 +262,8 @@
        */
       _minDate: {
         type: Date,
-        value: null
+        // null does not work here because minimizer passes undefined to overlay (#351)
+        value: ''
       },
 
       /**
@@ -270,7 +271,7 @@
        */
       _maxDate: {
         type: Date,
-        value: null
+        value: ''
       },
 
       _noInput: {
@@ -394,7 +395,7 @@
 
     _handleDateChange: function(property, value, oldValue) {
       if (!value) {
-        this[property] = null;
+        this[property] = '';
         return;
       }
 
--- a/vaadin-date-picker-behavior.html
+++ b/vaadin-date-picker-behavior.html
@@ -404,10 +404,7 @@
         this.value = oldValue;
         return;
       }
-      if (!(this[property] instanceof Date) ||
-        date.getFullYear() !== this[property].getFullYear() ||
-        date.getMonth() !== this[property].getMonth() ||
-        date.getDate() !== this[property].getDate()) {
+      if (!vaadin.elements.datepicker.DatePickerHelper._dateEquals(this[property], date)) {
         this[property] = date;
       }
     },
--- a/vaadin-date-picker-helper.html
+++ b/vaadin-date-picker-helper.html
@@ -44,13 +44,10 @@
      * @return {Boolean} True if the given date objects refer to the same date
      */
     _dateEquals: function(date1, date2) {
-      if (date1 && date2) {
-        return date1.getFullYear() === date2.getFullYear() &&
+      return date1 instanceof Date && date2 instanceof Date &&
+          date1.getFullYear() === date2.getFullYear() &&
           date1.getMonth() === date2.getMonth() &&
           date1.getDate() === date2.getDate();
-      } else {
-        return false;
-      }
     },
 
     /**


tomivirkki:
*[vaadin-date-picker-behavior.html, line 266 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/352#-KhaeYLJzLyf6ReisJv8:-KhaeYLJzLyf6ReisJv9:bvf7iru) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/2371d1987205a1644d998ab0548ba20ee8dc2ec2/vaadin-date-picker-behavior.html#L266)):*
> ```HTML
>         type: Date,
>         // null does not work here because minimizer passes undefined to overlay (#351)
>         value: new Date(-8640000000000000)
> ```

I think this will share the same date instance between every date picker on the page which sounds scary.

Doesn't the following work here as a fix?
```
value: function() {
  return null;
}
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/352)*

manolo:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

*[vaadin-date-picker-behavior.html, line 266 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/352#-KhaeYLJzLyf6ReisJv8:-Khaf7AbaCbBqP_WV30e:b-xg7nw1) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/2371d1987205a1644d998ab0548ba20ee8dc2ec2/vaadin-date-picker-behavior.html#L266)):*

I think this will share the same date instance between every date picker on the page which sounds scary.

Doesn't the following work here as a fix?
```
value: function() {
  return null;
}
```
</blockquote></details>

Already tested that and does not fixes

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/352)*

tomivirkki:
*[vaadin-date-picker-behavior.html, line 266 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/352#-KhaeYLJzLyf6ReisJv8:-KhajBWoRZ2Wip2KMoH-:b12c1kd) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/2371d1987205a1644d998ab0548ba20ee8dc2ec2/vaadin-date-picker-behavior.html#L266)):*

Already tested that and does not fixes
</blockquote></details>

Ok. Can you still define the value in a function as adviced here: https://www.polymer-project.org/1.0/docs/devguide/properties#configure-values

"When initializing a property to an object or array value, use a function to ensure that each element gets its own copy of the value, rather than having an object or array shared across all instances of the element."

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/352)*

manolo:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion, some commit checks broke.

---

*[vaadin-date-picker-behavior.html, line 266 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/352#-KhaeYLJzLyf6ReisJv8:-KhaofbULuThxKw4nFdx:bbyum0t) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/2371d1987205a1644d998ab0548ba20ee8dc2ec2/vaadin-date-picker-behavior.html#L266)):*

Ok. Can you still define the value in a function as adviced here: https://www.polymer-project.org/1.0/docs/devguide/properties#configure-values

"When initializing a property to an object or array value, use a function to ensure that each element gets its own copy of the value, rather than having an object or array shared across all instances of the element."
</blockquote></details>

That is not relevant for this case. Setting values to the first and the last valid dates does not have mayor problems when different instances share those dates. The property would be individually set per instance when the user changes the parameters, and the initial date would not be reused, so other instances are not modified.

Thus, the function does not fix the problem. If you prefer that I set the default values using a function I'll do but does not changes anything, only in the case that the date would be reused and changed internally. That is no happening.

The problem fixed here is just because values passed up to down from date-picker to the month-calendar, are received as undefined when it was null in origin. That is happening in some point of the polymer data binding code when it's optimized, so we prevent the issue just setting valid date values for the limits.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/352)*

tomivirkki:
*[vaadin-date-picker-behavior.html, line 266 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/352#-KhaeYLJzLyf6ReisJv8:-KharzPgjv9_xiw-d5yy:bi16jp7) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/2371d1987205a1644d998ab0548ba20ee8dc2ec2/vaadin-date-picker-behavior.html#L266)):*

That is not relevant for this case. Setting values to the first and the last valid dates does not have mayor problems when different instances share those dates. The property would be individually set per instance when the user changes the parameters, and the initial date would not be reused, so other instances are not modified.

Thus, the function does not fix the problem. If you prefer that I set the default values using a function I'll do but does not changes anything, only in the case that the date would be reused and changed internally. That is no happening.

The problem fixed here is just because values passed up to down from date-picker to the month-calendar, are received as undefined when it was null in origin. That is happening in some point of the polymer data binding code when it's optimized, so we prevent the issue just setting valid date values for the limits.


</blockquote></details>

I know it doesn’t change anything technically in this particular case but it’s a matter of (documented) convention. No problem keeping it as it is if you prefer this :)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/352)*

manolo:
Review status: 0 of 1 files reviewed at latest revision, all discussions resolved, some commit checks broke.

---

*[vaadin-date-picker-behavior.html, line 266 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/352#-KhaeYLJzLyf6ReisJv8:-Khaz2U1oOgvtLyBMU0h:b-896fix) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/2371d1987205a1644d998ab0548ba20ee8dc2ec2/vaadin-date-picker-behavior.html#L266)):*

I know it doesn’t change anything technically in this particular case but it’s a matter of (documented) convention. No problem keeping it as it is if you prefer this :)
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/352)*

tomivirkki:
*[vaadin-date-picker-behavior.html, line 266 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/352#-Ki0AtUdXVyXz87ORZ4u:-Ki0AtUdXVyXz87ORZ4v:bu4jc) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/e31c8b5476064cba66c31fa0bd8dd54130edeb99/vaadin-date-picker-behavior.html#L266)):*
> ```HTML
>         type: Date,
>         // null does not work here because minimizer passes undefined to overlay (#351)
>         value: ''
> ```

I'm glad that the fix works but not too happy that the fix is to set a _String_ literal as the default value for a type _Date_ property :( But if there's no other way around the issue then :lgtm:

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/352)*

manolo:
Review status: 0 of 3 files reviewed at latest revision, 1 unresolved discussion.

---

*[vaadin-date-picker-behavior.html, line 266 at r2](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/352#-Ki0AtUdXVyXz87ORZ4u:-Ki0GlFaSmzbpt0W_kzo:b1ve9zf) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/e31c8b5476064cba66c31fa0bd8dd54130edeb99/vaadin-date-picker-behavior.html#L266)):*

I'm glad that the fix works but not too happy that the fix is to set a _String_ literal as the default value for a type _Date_ property :( But if there's no other way around the issue then :lgtm:
</blockquote></details>

Well, it's the simpler approach.
We could use pure Date objects to set the min and max default values, and assign these default values when attributes are not set, or when they are assigned to empty/null/undefined strings. That needs more LOC and doesn't leave a clearer code.
I guess, this will not be an issue with Polymer 2, so it's a temporary hack in the meanwhile we continue supporting 1.x.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/352)*

platosha:
:man_facepalming: :lgtm:

---

Reviewed 3 of 3 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/352#-:-Ki4PNG8dewaMPcEec4e:bg6m2tn)*

PR #361, opened 2017-05-03
Split SauceLabs job into 2 jobs, make biuld process more readable

### LINKED ISSUE #357: Fix CRON build - two of seven platforms are failing because of SauceLabs limit
Either:

- increase SauceLabs timeout if it's possible (most probably it's not possible)
- split the CRON SauceLabs job into 2 jobs: 1(iphone+ipad+edge+ie+safari) + 2(win/chrome+win/firefox)

limonte:
Fixes #357 

- CRON build splitted to 2 parts: regular SauceLabs build (5 platforms) + CRON platforms (win/chrome and win/firefox) 
- `.travis.yml` and `wct.conf.js` are now less compex and easier to read
- all npm deps bumbed to latest
- switch to yarn, it's faster

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-date-picker/361)


tomivirkki:
iPhone and iPad tests logged "Unexpected error. { Error: socket hang up" at https://travis-ci.org/vaadin/vaadin-date-picker/builds/228340008?utm_source=github_status&utm_medium=notification

Is this something we should worry about?

limonte:
> iPhone and iPad tests logged "Unexpected error. { Error: socket hang up" at https://travis-ci.org/vaadin/vaadin-date-picker/builds/228340008?utm_source=github_status&utm_medium=notification
> 
> Is this something we should worry about?

I don't think we should worry about it since the build is green. Also, this warning is outside the scope of this PR, it was here before: https://travis-ci.org/vaadin/vaadin-date-picker/builds/218837756

manolo:
Review status: 0 of 4 files reviewed at latest revision, 1 unresolved discussion, some commit checks broke.

---

*[.travis.yml, line 12 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/361#-KjcffZ0lHsbizaUabZQ:-KjcffZ0lHsbizaUabZR:b4e8yic) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/536ace1667393f65c500958a64358bf46f2741d1/.travis.yml#L12)):*
> ```YAML
> 
> addons:
>   apt:
> ```

Why do you install yarn as a debian package instead of adding that dev-dependency and install with npm ? 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/361)*

limonte:
Review status: 0 of 4 files reviewed at latest revision, 1 unresolved discussion, some commit checks broke.

---

*[.travis.yml, line 12 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/361#-KjcffZ0lHsbizaUabZQ:-Kjgp-qUlJqVznzEjwjm:b-77p9ul) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/536ace1667393f65c500958a64358bf46f2741d1/.travis.yml#L12)):*

Why do you install yarn as a debian package instead of adding that dev-dependency and install with npm ? 
</blockquote></details>

The whole point of this "npm -> yarn" change is to get rid of npm, because it takes twice more than yarn disk space and it's slower. As you can notice there's no `npm install` at all anymore.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/361)*

PR #413, opened 2017-08-11
Fix invalid property to reflect correct value on value-changed

### LINKED ISSUE #410: The `invalid` property doesn't reflect the state of the component during a `value-changed` event.
When you add a `value-changed` listener to the component, you always receive an event whenever the value has changed - even if the value is invalid.

So for example, if you create a `vaadin-date-picker` with min a max dates, like this:
```
<vaadin-date-picker id="my-picker" min="2017-08-01" max="2017-08-31">
</vaadin-date-picker>
```
... and then put a listener on it:
```
let picker = document.getElementById('my-picker');
picker.addEventListener('value-changed', (e) => console.log('Value changed to: ' + e.detail.value + ' invalid is ' + e.target.invalid));
```
... and then select a valid date (01/08/2017), the message is:
```
Value changed to: 2017-08-01 invalid is false
```
But if you type an invalid date, like 01/08/2018, the `invalid` flag is still `false`:
```
Value changed to: 2018-08-01 invalid is false
```
And then if you return to a valid value, the flag is `true`:
```
Value changed to: 2017-08-01 invalid is true
```

So the problem is: the `invalid` property is only changed after the value has changed. So I can't verify if the changed value is valid or not by looking at the property.

### Live demo

https://jsbin.com/foziloqudi/edit?html,js,console,output
(You have to execute the steps manually)

samiheikki:
Fixes #410

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-date-picker/413)

### DIFF
--- a/test/basic.html
+++ b/test/basic.html
@@ -501,6 +501,18 @@
           expect(datepicker.invalid).to.be.equal(false);
         });
 
+        it('should reflect correct invalid value on value-changed eventListener', done => {
+          datepicker.value = '2016-01-01'; // valid
+
+          datepicker.addEventListener('value-changed', () => {
+            expect(datepicker.invalid).to.be.equal(true);
+            done();
+          });
+
+          datepicker.open();
+          datepicker.$.overlay.selectedDate = new Date('2017-01-01'); // invalid
+        });
+
         it('should scroll to min date when today is not allowed', done => {
           datepicker.max = null;
           datepicker.min = '2100-01-01';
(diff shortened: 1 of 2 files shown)


tomivirkki:
:lgtm_strong: 

---

Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/413#-:-KrFT6AMPHeahKFjDOnJ:b-uwkowd)*

Issue #421, opened 2017-09-01
Horizontal Alignment calculation is wrong in RTL mode


coderek:
### Description

`vaadin-date-picker` is right aligned erroneously in RTL mode, resulting in squashed view. 

### Expected outcome

It should use the provided space smartly and align to the left when there is enough space in the right in RTL mode.

### Live Demo

https://drive.google.com/file/d/0BxQL_fXYFDfiR2JNT0xXOHMzVTg/view?usp=sharing

### Steps to reproduce

1. put `dir=[rtl]` to body tag
2. open a page with `vaadin-date-picker`, it should be "right" aligned (left in rtl)
3. resize the window to reduce the space on the left of the input box
4. the horizontal alignment is changed and the `vaadin-date-picker` becomes shrank 

### Browsers Affected

- [x] Chrome
- [x] Firefox
- [x] Safari
- [x] Edge
- [x] IE 11
- [x] iOS Safari
- [x] Android Chrome

PR #422, opened 2017-09-02
Fix alignment in rtl mode

coderek:
Fixes https://github.com/vaadin/vaadin-date-picker/issues/421

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-date-picker/422)

### DIFF
--- a/vaadin-date-picker-mixin.html
+++ b/vaadin-date-picker-mixin.html
@@ -336,6 +336,9 @@ This program is available under Apache License Version 2.0, available at https:/
       this.addEventListener('date-tap', this._close.bind(this));
       this.addEventListener('focus-input', this._focusAndSelect.bind(this));
       this.addEventListener('input', this._onUserInput.bind(this));
+
+      const rtlEl = document.querySelector('[dir=rtl]');
+      this._isRtl = rtlEl && rtlEl.contains(this);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


coderek:
We are actually using vaadin-date-picker v1.2.3 right now. Do I need to create another pull request to merge into older version?

vaadin-date-picker-mixin.html
```diff
@@ -10,6 +10,7 @@
 
 <script>
   window.Vaadin = window.Vaadin || {};
+  var isRtl = !!document.querySelector('[dir=rtl]');
```
jouni: Should this somehow check that it actually affects the date picker element? I mean, in theory, it could be set in a part of the HTML tree which does not contain the date picker. But I suppose that’s very theoretical worry, since usually it’s the whole page that is ltr/rtl.

coderek: I've changed accordingly to check if the picker is inside a `[dir=rtl]` element.

coderek:
@jouni Hi Jouni,
We are using date-picker version 1.2.3. Do you guys still update v1?

jouni:
Sorry for the awfully long delay in response! 😳

> We are using date-picker version 1.2.3. Do you guys still update v1?

Right now we are in a really tight crunch mode, trying to get the next major versions of all elements done, and some new elements altogether. So right now the previous versions are not getting much love, unfortunately.

I hope we can dedicate more time for maintenance later this year, but I can’t guarantee much in the near future. I’ll ping the team about this in any case.

platosha:
Hi, thanks for pointing out this issue.

Unfortunately we are not adding features to v1 of the `<vaadin-date-picker>` anymore.

Our plan is to use `<vaadin-overlay>` in the date picker in future. This should resolve the issue with incorrect `horizontalAlign` too.

PR #427, opened 2017-09-11
Support Firefox in webcomponents.org

limonte:
Related to https://github.com/vaadin/vaadin-element-skeleton/pull/58

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-date-picker/427)

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,7 @@
 ```
 <custom-element-demo height="550">
   <template>
+    <script src="../webcomponentsjs/webcomponents-lite.js"></script>
     <link rel="import" href="vaadin-date-picker.html">
     <next-code-block></next-code-block>
   </template>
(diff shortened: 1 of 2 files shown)


tomivirkki:
Why does `<paper-input>` for example work without this hack?

limonte:
Because of https://github.com/webcomponents/webcomponents.org/blob/04ba2624ff284d18a98297eb24e6425827c5d023/client/bower.json#L22

We need `webcomponentsjs 1.x` for our components.

tomivirkki:
Yeah, looks like it's lacking from date-picker devDependencies https://github.com/vaadin/vaadin-date-picker/blob/master/bower.json even though it's directly used in the demos

limonte:
Added `webcomponentsjs` devDependency, it works now: https://www.webcomponents.org/preview/vaadin/vaadin-date-picker/5dbb679233e560c5a079e8115355741608b52418

tomivirkki:
LGTM

PR #449, opened 2017-10-14
Allow setting date/min/max using date objects. Fix #446

### LINKED ISSUE #446: [NFR] Allow value to be set with a Date object
Allowing a user to set `value` as a standard `Date` object would make programatic work with the element _MUCH_ easier. When working with JS, setting a `value` can get rather cumbersome as there are no built-in methods that output strictly the ISO 8601 date segment. Right now I'm having to do something like:
```
datePicker.value = new Date().toISOString().split('T')[0];
```
to set the default in browser.

manolo:
Fixes #446

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-date-picker/449)

### DIFF
--- a/test/basic.html
+++ b/test/basic.html
@@ -54,6 +54,16 @@
         expect(datepicker.value).to.equal('');
       });
 
+      it('should be possible to set value by using strings', () => {
+        datepicker.value = '2013-11-02';
+        expect(datepicker.value).to.equal('2013-11-02');
+      });
+
+      it('should be possible to set value by using date objects', () => {
+        datepicker.value = new Date(2017, 1, 11);
+        expect(datepicker.value).to.equal('2017-02-11');
+      });
+
       // Needs to be in this order to pass in IE
       it('should not show clear icon', done => {
         datepicker.$.overlay.$.monthScroller.bufferSize = 1;
--- a/vaadin-date-picker-mixin.html
+++ b/vaadin-date-picker-mixin.html
@@ -429,7 +429,7 @@ This program is available under Apache License Version 2.0, available at https:/
         return;
       }
 
-      var date = this._parseDate(value);
+      var date = value instanceof Date ? value : this._parseDate(value);
       if (!date) {
         this.value = oldValue;
         return;


tomivirkki:
*[vaadin-date-picker-mixin.html, line 432 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/449#-KwTv3iWJaPE7VTPxby9:-KwTv3iWJaPE7VTPxbyA:b-48nth2) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/6f0bf513ba91d32219f0a4fb2e6c1632e99133ab/vaadin-date-picker-mixin.html#L432)):*
> ```HTML
>       }
> 
>       var date = value instanceof Date ? value : this._parseDate(value);
> ```

This is not ok. Value is of [type String](https://github.com/vaadin/vaadin-date-picker/blob/master/vaadin-date-picker-mixin.html#L42) (also documented so) and we're not changing that to strangely accept other data types. An additional utility function could possibly be added instead, something like `setValueByDate(date)`...

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/449)*

manolo:
Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion.

---

*[vaadin-date-picker-mixin.html, line 432 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/449#-KwTv3iWJaPE7VTPxby9:-KwZb20oQruXt1ig52yL:b-w1pe1q) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/6f0bf513ba91d32219f0a4fb2e6c1632e99133ab/vaadin-date-picker-mixin.html#L432)):*

This is not ok. Value is of [type String](https://github.com/vaadin/vaadin-date-picker/blob/master/vaadin-date-picker-mixin.html#L42) (also documented so) and we're not changing that to strangely accept other data types. An additional utility function could possibly be added instead, something like `setValueByDate(date)`...
</blockquote></details>

Agree that this need to be documented like in public/bower_components/iron-autogrow-textarea/iron-autogrow-textarea.html 
But I think a new method is more work and does not facilitate easy databinding for the final user.

I would rather

```
  /**
   * ...
   * @type {String|Date}
   */
  value: {
    type: Object, // (or String since double type is not allowed here)
    observer: '_valueChanged',
    notify: true,
    value: ''
  },,
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/449)*

tomivirkki:
*[vaadin-date-picker-mixin.html, line 432 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/449#-KwTv3iWJaPE7VTPxby9:-Kw_NCsTosGi0dJafbAN:b-n8xxjp) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/6f0bf513ba91d32219f0a4fb2e6c1632e99133ab/vaadin-date-picker-mixin.html#L432)):*

Agree that this need to be documented like in public/bower_components/iron-autogrow-textarea/iron-autogrow-textarea.html 
But I think a new method is more work and does not facilitate easy databinding for the final user.

I would rather

```
  /**
   * ...
   * @type {String|Date}
   */
  value: {
    type: Object, // (or String since double type is not allowed here)
    observer: '_valueChanged',
    notify: true,
    value: ''
  },,
```
</blockquote></details>

So you can set the value as a Date object, but still when a user selects a value from the dropdown, it's suddenly overridden by a String value. Doesn't sound like it facilitates easy data binding when the type is unpredictable.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/449)*

platosha:
Reviewed 2 of 2 files at r1.

---

*[vaadin-date-picker-mixin.html, line 432 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/449#-KwTv3iWJaPE7VTPxby9:-KweNjbEL3kWfAUPTcED:b-12thj2) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/6f0bf513ba91d32219f0a4fb2e6c1632e99133ab/vaadin-date-picker-mixin.html#L432)):*

So you can set the value as a Date object, but still when a user selects a value from the dropdown, it's suddenly overridden by a String value. Doesn't sound like it facilitates easy data binding when the type is unpredictable.
</blockquote></details>

Agree with @tomivirkki on that, accepting multiple value formats on the same property is confusing.

FYI the datepicker mixin has a `_formatISO(date)` helper method, which does the conversion job in the right way. We probably just need to expose it as a public method:

```javascript
datePicker.value = datePicker.formatISODate(new Date());
```

Maybe even as a static method, as there is nothing instance-specific in the conversion method:

```javascript
datePicker.value = Vaadin.DatePickerElement.formatISODate(new Date());
```

This could help in computed data bindings as well.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/449)*

manolo:

---

*[vaadin-date-picker-mixin.html, line 432 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/449#-KwTv3iWJaPE7VTPxby9:-Kwtja84HUkgKTE8pE_j:b-owz9jb) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/6f0bf513ba91d32219f0a4fb2e6c1632e99133ab/vaadin-date-picker-mixin.html#L432)):*
<details><summary><i>Previously, platosha (Anton Platonov) wrote…</i></summary><blockquote>

Agree with @tomivirkki on that, accepting multiple value formats on the same property is confusing.

FYI the datepicker mixin has a `_formatISO(date)` helper method, which does the conversion job in the right way. We probably just need to expose it as a public method:

```javascript
datePicker.value = datePicker.formatISODate(new Date());
```

Maybe even as a static method, as there is nothing instance-specific in the conversion method:

```javascript
datePicker.value = Vaadin.DatePickerElement.formatISODate(new Date());
```

This could help in computed data bindings as well.
</blockquote></details>

Another option could be to have a `date` that reflects the value and can be modifiable so as the user can bind either to `value` or `date` 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/449)*

platosha:

---

*[vaadin-date-picker-mixin.html, line 432 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/449#-KwTv3iWJaPE7VTPxby9:-KxNP4MyITFFnxbuaHxo:bla8lsm) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/6f0bf513ba91d32219f0a4fb2e6c1632e99133ab/vaadin-date-picker-mixin.html#L432)):*

Another option could be to have a `date` that reflects the value and can be modifiable so as the user can bind either to `value` or `date` 
</blockquote></details>

Not a good option IMO. It is hard to detect mutations on Date instances, consider: `datePicker.date = new Date(); datePicker.date.setDate(1);`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/449)*

manolo:

---

*[vaadin-date-picker-mixin.html, line 432 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/449#-KwTv3iWJaPE7VTPxby9:-KxYKkxHcq0U7AwyD-Cb:bbp83gu) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/6f0bf513ba91d32219f0a4fb2e6c1632e99133ab/vaadin-date-picker-mixin.html#L432)):*
<details><summary><i>Previously, platosha (Anton Platonov) wrote…</i></summary><blockquote>

Not a good option IMO. It is hard to detect mutations on Date instances, consider: `datePicker.date = new Date(); datePicker.date.setDate(1);`.
</blockquote></details>

Polymer users already are familiar with data-binding provided by the library and it's limitations, hence how the user binds to a date object is something we shouldn't care. 

IMO, It's much more important to facilitate the user the interaction with the component as the reporter manifests in #446. I don't think the continuous call to `formatISODate` is a good option from the point of view of simplicity (ffs) or performance (not a big deal but counts)

Even vaadin-date-picker already offers API for date objects in the public API of `<vaadin-month-calendar>` (focusedDate, min and max).

Anyways, what should I do here, abandon this PR and close the #446 issue, or propose a solution based on a `date` property?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/449)*

platosha:

---

*[vaadin-date-picker-mixin.html, line 432 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/449#-KwTv3iWJaPE7VTPxby9:-KxgR6NzLpwEKU4hWiGv:bz6a1dt) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/6f0bf513ba91d32219f0a4fb2e6c1632e99133ab/vaadin-date-picker-mixin.html#L432)):*

Polymer users already are familiar with data-binding provided by the library and it's limitations, hence how the user binds to a date object is something we shouldn't care. 

IMO, It's much more important to facilitate the user the interaction with the component as the reporter manifests in #446. I don't think the continuous call to `formatISODate` is a good option from the point of view of simplicity (ffs) or performance (not a big deal but counts)

Even vaadin-date-picker already offers API for date objects in the public API of `<vaadin-month-calendar>` (focusedDate, min and max).

Anyways, what should I do here, abandon this PR and close the #446 issue, or propose a solution based on a `date` property?
</blockquote></details>

My suggestion here stays same as before:

- not introduce another accepted property type
- not introduce another property to keep in sync
- introduce a static helper function for `Date` instance → ISO-formatted local date `String` conversion. Note: use the date picker’s implementation, as builtin `date.toISOString()` brings timezone issues.

I see what you mean, it is indeed not the best DX to have to convert a `Date` to `String` every time. However, I strongly feel this is not in our hands to fix the world at this point.

The thing is that JS `Date` object also stores extra time with timezone, what the date picker alone does not take and does not give as a result. Back in the days of first implementation, there were lots of issues in the date picker coming from that extra time being an obstacle and influencing the date itself in a number of ways. We found our workarounds to use inside the date picker, but they are not necessarily working for everyone. I feel that if we expose value as `Date`, we would bring all these extra time issues to our users then. I wouldn’t like this to happen. IMO, the best we can do about extra time is not to deal with extra time at all, and hence, not use the `Date`.

In a datetime picker element, though, having `Date` support would make perfect sense.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/449)*

manolo:
Apparently we are not going to provide date objects. So closing this.

PR #462, opened 2017-11-07
Fix custom validator demo

### LINKED ISSUE #460: Custom validator demo broken
The "custom validator" demo in the Advanced sections is broken:

demo

Actually, there are two bugs here:

1. value argument isn't passed to `super.checkValidity()` call in [line 68](https://github.com/vaadin/vaadin-date-picker/blob/master/demo/date-picker-advanced-demos.html#L68);

2. having this fixed, another bug reveals: once invalid, with the given snippet the element can't become valid anymore due to `super.checkValidity` call falling per invoking `checkValidity` on the `vaadin-text-field` element, especially [line 219](https://github.com/vaadin/vaadin-text-field/blob/master/vaadin-text-field-mixin.html#L219).

Other than that, the code snippet used in demo is not aligned with the [form-input](https://github.com/vaadin/vaadin-date-picker/blob/master/test/form-input.html) test which doesn't call `super.checkValidity` at all.

web-padawan:
Fixes #460

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-date-picker/462)

### DIFF
--- a/demo/date-picker-advanced-demos.html
+++ b/demo/date-picker-advanced-demos.html
@@ -65,12 +65,8 @@
           window.addDemoReadyListener('#date-picker-advanced-demos-custom-validator', function(document) {
             class DatePickerCurrentYearElement extends Vaadin.DatePickerElement {
               checkValidity(value) {
-                var valid = super.checkValidity();
-                if (valid) {
-                  var currentYear = new Date().getFullYear();
-                  valid = new Date(value).getFullYear() === currentYear;
-                }
-                return valid;
+                var currentYear = new Date().getFullYear();
+                return new Date(value).getFullYear() === currentYear;
               }
             }
             customElements.define('vaadin-date-picker-current-year', DatePickerCurrentYearElement);


web-padawan:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/462#-KyL7Nivl2oslwHtlaTp:-KyL7Nivl2oslwHtlaTq:bkzohr6) (no related file):*
I was wondering if `super.checkValidity()` call should be supposed here?
The previous demo was broken due to the fact described in issue: once invalid value is selected, the subsequent `checkValidity` call for text field fails even after selecting valid value.

After this fix, the demo is aligned with the test. But it seems this might require additional test. @manolo WDYT?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/462)*

manolo:
Reviewed 1 of 1 files at r1.

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/462#-KyL7Nivl2oslwHtlaTp:-KyZwv8IyFNjEvhuOYFr:b607pic) (no related file):*
<details><summary><i>Previously, web-padawan (Sergey Kulikov) wrote…</i></summary><blockquote>

I was wondering if `super.checkValidity()` call should be supposed here?
The previous demo was broken due to the fact described in issue: once invalid value is selected, the subsequent `checkValidity` call for text field fails even after selecting valid value.

After this fix, the demo is aligned with the test. But it seems this might require additional test. @manolo WDYT?
</blockquote></details>

The objective of calling super was to make default validator run other custom validation conditions,  like required, min, max (btw there is no pattern in this component). For demo purposes is simpler not to call super.

:lgtm: 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/462)*

PR #469, opened 2017-11-12
Es5 demos

samiheikki:
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-date-picker/469)

### DIFF
--- a/demo/date-picker-advanced-demos.html
+++ b/demo/date-picker-advanced-demos.html
@@ -40,7 +40,7 @@
               },
               parseDate: function(dateString) {
                 var matches = datepicker.i18n.monthNames.filter(function(monthName) {
-                  return monthName.toLowerCase().startsWith(dateString.trim().toLowerCase());
+                  return monthName.toLowerCase().substr(0, dateString.trim().toLowerCase().length) === dateString.trim().toLowerCase();
                 });
                 dateString = matches.length === 1 ? matches[0] : dateString;
                 // Parse the date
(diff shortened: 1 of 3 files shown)


tomivirkki:
*[demo/date-picker-validators-demos.html, line 37 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/469#-Kz3SCO2NcRCtNqzUDMJ:-Kz3SCO2NcRCtNqzUDMK:buitt9k) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/4de45890b9b77e9f131b80fd93344c0bb5a07562/demo/date-picker-validators-demos.html#L37)):*
> ```HTML
>       }
>       static get template() {
>         if (Array.prototype.find) {
> ```

Hmm..for some reason this doesn't work (even though the same approach works in text-field demos). Maybe need to check user agent (for IE) instead.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/469)*

samiheikki:
Review status: 0 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[demo/date-picker-validators-demos.html, line 37 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/469#-Kz3SCO2NcRCtNqzUDMJ:-Kz4SZO82UhBrebGKDEz:bdiwazw) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/4de45890b9b77e9f131b80fd93344c0bb5a07562/demo/date-picker-validators-demos.html#L37)):*

Hmm..for some reason this doesn't work (even though the same approach works in text-field demos). Maybe need to check user agent (for IE) instead.
</blockquote></details>

That's right, it seems to be polyfilled somewhere. I've thought that we could rather check that is this feature polyfilled or native code. This works at the moment, but will break if some date-picker dependency is polyfilling that feature even tough it is natively supported. Let me know what you think about this. After deciding I can make the same change to text-field demo also.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/469)*

tomivirkki:
*[demo/date-picker-validators-demos.html, line 37 at r1](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/469#-Kz3SCO2NcRCtNqzUDMJ:-KzY3zKGTExhprsLGnFg:b-w3iwec) ([raw file](https://github.com/vaadin/vaadin-date-picker/blob/4de45890b9b77e9f131b80fd93344c0bb5a07562/demo/date-picker-validators-demos.html#L37)):*
> Array.prototype.find.toString()

Doesn't this error if `Array.prototype.find` isn't defined? Would this check do the trick https://stackoverflow.com/a/30692705 ?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/vaadin-date-picker/469)*

web-padawan:
Closing in favour of #608

PR #575, opened 2018-06-11
Add test folder to bower ignores

tomalec:
To make package slightly smaller

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-date-picker/575)

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -28,7 +28,8 @@
     "bower_components",
     "gulpfile.js",
     "package-lock.json",
-    "wct.conf.js"
+    "wct.conf.js",
+    "test"
   ],
   "dependencies": {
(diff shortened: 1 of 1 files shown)


web-padawan:
Thank you for the PR. We keep tests published to bower intentionally, to run  these tests in `vaadin-overlay` and `vaadin-control-state-mixin` for all the components depending on them.

See the example here: https://github.com/vaadin/vaadin-overlay/blob/master/test/index.html

Furthermore, AFAIK ignoring has no effect: unlike npm, bower still downloads zip package from GitHub, then extracts all the files except those ignored. 

So I have to close this PR, but thank you again for all your contributions! We appreciate it.

PR #610, opened 2018-09-30
Fix aria role for calendar dates

samiheikki:
Should use `presentation` role for calendar dates: https://www.w3.org/TR/2017/WD-wai-aria-practices-1.1-20170628/examples/presentation/PresentationRoleExamples.html

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-date-picker/610)


samiheikki:
Not sure but I'll take a look

PR #622, opened 2018-12-07
Live demo and API Documentation into own section.

dwkns:
The live demo and API documentation links get lost at the top of the document. The first expectations in *most* GitHub repositories is an explanation not links somewhere else. 

Also these are important links and not easy to find when scanning the page. Putting them in their own section improves findability.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-date-picker/622)

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,10 +10,6 @@
 
 # &lt;vaadin-date-picker&gt;
 
-[Live Demo ↗](https://vaadin.com/components/vaadin-date-picker/html-examples)
-|
-[API documentation ↗](https://vaadin.com/components/vaadin-date-picker/html-api)
-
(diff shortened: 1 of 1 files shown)


web-padawan:
Thanks for the PR. We generally follow the same concepts as Polymer elements, see e.g.
https://github.com/PolymerElements/paper-button

I agree that description could be the first line, and the links could go after it, but not in the separate section (as it would be eventually more difficult to discover).

Also, please take the following into account when contributing:

1. Any changes to readme, like this one, should be first submitted to the skeleton repo:
https://github.com/vaadin/vaadin-element-skeleton
Once we discuss it there and decide if we should merge or decline, we will apply change to components.

2. Any changes to readme and demos should have `[skip ci]` in the end of the commit message in order to prevent triggering Travis build for nothing.

web-padawan:
Closing in favour of vaadin/vaadin-element-skeleton#186

PR #637, opened 2019-02-22
Switch from dom-if to disable-upgrade and update tests

### LINKED ISSUE #635: Tests on IE11 are broken
Tests are broken on IE11 after introducing performance improvement: https://github.com/vaadin/vaadin-date-picker/commit/2d86da5dd1a92fdd9c1d39b6562c36e528879897#diff-0f7cab7f0a6b1e79bf518eb211b482deR328

### LINKED ISSUE #638: A date picker that has a value before it's opened doesn't open on IE11
The following date picker doesn't open on IE11
```
<vaadin-date-picker label="Birthday" value="1980-08-14">
</vaadin-date-picker>
```

yuriy-fix:
Fixes #635
Fixes #638 

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-date-picker/637)

### DIFF
--- a/src/vaadin-date-picker-mixin.html
+++ b/src/vaadin-date-picker-mixin.html
@@ -321,11 +321,8 @@ This program is available under Apache License Version 2.0, available at https:/
     }
 
     _initOverlay() {
-      const template = this.shadowRoot.querySelector('#overlay-template');
-      template.if = true;
-      template.render();
+      this.$.overlay.removeAttribute('disable-upgrade');
 
-      this.$.overlay = this.shadowRoot.querySelector('#overlay');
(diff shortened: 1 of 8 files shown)


src/vaadin-date-picker-mixin.html
```diff
@@ -344,7 +341,7 @@
     disconnectedCallback() {
       super.disconnectedCallback();
 
-      if (this.$.overlay) {
+      if (!this.$.overlay.hasAttribute('disable-upgrade')) {
```
web-padawan: consider adding a property, e.g. `_overlayInitialized` and using it instead

yuriy-fix: Updated.

PR #648, opened 2019-03-20
feat: add clear-button-visible API from <vaadin-text-field>

### LINKED ISSUE #645: Integrate text-field's clearButtonVisible API
Remove the existing dedicated clear button and use the included text-field's clearButtonVisible API instead.
Fix the tests to use the new functionality.
Add demo, update visual tests...

The new functionality will make the clear button hidden by default = breaking change

platosha:
Fixes #645

BREAKING CHANGE: The clear button is now hidden by default. To make it
visible, set the `clear-button-visible` attribute or use
`clearButtonVisible` JS API property.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-date-picker/648)


Haprog:
Should probably also bump `vaadin-text-field` dependency version in `bower.json` from `^2.1.2` to `^2.3.0` since [v2.3.0](https://github.com/vaadin/vaadin-text-field/releases/tag/v2.3.0) introduces the `clear-button-visible` API.

platosha:
@Haprog done

Haprog:
@platosha Looks like in this version when you click on the clear button, it always opens the date picker, regardless of if the date picker input was focused or not and if the overlay was open or not. This is different from current release version.

If you test it here: https://cdn.vaadin.com/vaadin-date-picker/3.3.2/demo/#date-picker-basic-demos clicking on the clear button doesn't open or close the date-picker, it just clears the value and focuses the text field. So if you click the clear button when the overlay is open, it stays open. And if you click the clear button when the overlay is closed, it stays closed.

We should probably add tests for this so that the behaviour doesn't unintentionally change. So we should test what happens when the clear button is clicked in at least these cases:

**Clear button is clicked when:**
1. date picker is not focused, overlay is closed
2. date picker input is focused, overlay is closed (e.g. right after selecting a date from the overlay which closes the overlay but leaves the text input focused)
3. date picker overlay is open

test/dropdown.html
```diff
@@ -157,24 +156,6 @@
         datepicker.opened = false;
         expect(datepicker.focusElement.hasAttribute('focus-ring')).to.be.false;
       });
-
-      it('should not close on clear-button down', () => {
```
Haprog: I think these two tests should be kept and the behaviour needs to be fixed so that these still pass. Now the behaviour is not so nice anymore, when overlay always opens. See my comment here: https://github.com/vaadin/vaadin-date-picker/pull/648#issuecomment-476639634

src/vaadin-date-picker.html
```diff
@@ -57,12 +57,12 @@
         disabled="[[disabled]]"
         readonly="[[readonly]]"
         error-message="[[errorMessage]]"
+        clear-button-visible="[[clearButtonVisible]]"
         aria-label$="[[label]]"
         part="text-field"
         theme$="[[theme]]"
       >
       <slot name="prefix" slot="prefix"></slot>
-      <div part="clear-button" slot="suffix" on-touchend="_clearTouchend" on-click="_clear" role="button" aria-label$="[[i18n.clear]]"></div>
```
Haprog: https://github.com/vaadin/vaadin-date-picker/blob/33cfb8457779177c6e7d5c57143f8002eb350b7d/src/vaadin-date-picker.html#L39

This CSS block should be removed since there is no more `part="clear-button"` in this template.

Haprog: There's also this clear button:  
https://github.com/vaadin/vaadin-date-picker/blob/33cfb8457779177c6e7d5c57143f8002eb350b7d/src/vaadin-date-picker-overlay-content.html#L165

which I think should be removed. It looks like that one is never shown. It also has some associated CSS which could then be removed (in the same template and also in `vaadin-date-picker-overlay-content-styles.html`).

It also looks like the `_clear()` methods in both `vaadin-date-picker.html` and also `vaadin-date-picker-overlay-content.html` are not used anywhere so those could probably be removed?

platosha: > which I think should be removed. It looks like that one is never shown.

This clear button is shown in the overlay in Material theme on mobile.

platosha: BTW here you can try a material date-picker: https://cdn.vaadin.com/vaadin-material-styles/1.2.2/demo/date-pickers.html

Haprog: Ok, I didn't notice it's only shown with Material theme. I just looked that with Lumo it was not shown in desktop or mobile. I think it's a bit weird that a control in a specific view is only shown in one theme. I wonder what's the reason it's not shown with Lumo. But I guess this is fine now. Thanks.

test/dropdown.html
Haprog: Latest version looks good now. Thanks!

Issue #663, opened 2019-06-05
Add support for date patterns


mhosio:
As a user I would like to use patterns such as "d.M.yyyy" for formatting the date. This functionality has been used by many on the earlier versions of the Vaadin Framework and it is still not yet available for the new date picker.

web-padawan:
Please check my comment here https://github.com/vaadin/vaadin-date-picker/issues/662#issuecomment-496513325

We already provide the `i18n` API which is enough to implement the custom parser and formatter.

Baking support for various patterns into `vaadin-date-picker` itself would make it more opinionated and less generic, and also it would result in growing bundle size (which is already quite big).

mhosio:
Now that I think of it, I agree with you. And the api makes it fairly easy to integrate third party date parsing libraries when pattern support is needed. This is more of an issue for the java-api (flow) where overriding i18n functions is trickier. However, there seems to be already an issue regarding that so I will close this one now. Thanks for the input.

Issue #662, opened 2019-05-28
Let DateField support an alternative date format


cdir:
### Description

In #661 is already described that we need some way to enter a date more quickly by using only two digit years. In business applications with many date input fields the user is used to enter the date without the date seperators. That seperators are created automatically or after leaving the field.

### Expected outcome
For example a date could be entered using just 6 digit. 010121 will be transformed to 01/01/2021. It may be even possible to enter the format dependent seperator automatically (dependent on the locale). In date format with day or month first, the seperator could be added automatically after entering two digits if there is not yet a seperator. An additional seperator will be automatically ignored.

A simpler approach may be to have the ability to give a primary and a list of alternative date formats. If the primary date format does not leads to a valid date, the alternative date formats are tried out until there is a valid date. The developer have to make sure that the order of the date format is useful. Afterwards the date is formatted in primary date format. For a german date picker I would use dd.MM.YYYY as primary format and ddMMYYYY as an alternative format. With #661 also two digit year should be supported this way.

web-padawan:
Use [`i18n` property](https://cdn.vaadin.com/vaadin-date-picker/4.0.1/#/elements/vaadin-date-picker#property-i18n) and especially `parseDate` and `formatDate` if you need a custom date format.
That way you could use any separator you want, as well as update it based on locale etc.

See also the [example](https://cdn.vaadin.com/vaadin-date-picker/4.0.1/demo/#date-picker-localization-demos) of using `i18n` property for localisation.

web-padawan:
Closing per above comment.

PR #665, opened 2019-06-13
Add option to show year as 2 digits

pierlon:
Relates to vaadin/vaadin-date-picker-flow#141.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-date-picker/665)

### DIFF
--- a/src/vaadin-date-picker-mixin.html
+++ b/src/vaadin-date-picker-mixin.html
@@ -156,9 +156,9 @@ This program is available under Apache License Version 2.0, available at https:/
               // A function to format given `Object` as
               // date string. Object is in the format `{ day: ..., month: ..., year: ... }`
               // Note: The argument month is 0-based. This means that January = 0 and December = 11.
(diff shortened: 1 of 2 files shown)


web-padawan:
Thanks for the PR. 

Overriding `i18n` API is a supposed way of handling such cases. But we will not bake any formats in the web component, as I previously explained in https://github.com/vaadin/vaadin-date-picker/issues/662#issuecomment-496513325 and https://github.com/vaadin/vaadin-date-picker/issues/663#issuecomment-499058619.

Please also note that there is a [parseDate](https://github.com/vaadin/vaadin-date-picker-flow/blob/1c87071ea3daf3b1b034665f3de7dcd576ed9ec2/vaadin-date-picker-flow/src/main/resources/META-INF/resources/frontend/datepickerConnector.js#L110) override in Flow counterpart. Maybe tweaking that could help.

pierlon:
Understood, thanks.

PR #679, opened 2019-10-03
Run validation when field is blurred

jouni:
Fixes https://github.com/vaadin/vaadin-date-picker-flow/issues/201

If you can come up with any potential side-effects of running validation on blur, that would be great to catch before merging 🤗

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-date-picker/679)


jouni:
The tests are failing at some random thing about “theme-propagation”. Doesn’t seem to be related to this change.

Test pass locally in Chrome at least.

jouni:
One point raised by the original issue author: pressing the clear button should remove the invalid state immediately as well. Right now it is not removed, only after the user blurs the field.

web-padawan:
> pressing the clear button should remove the invalid state immediately as well. Right now it is not removed, only after the user blurs the field.

Confirmed this issue. It works with `vaadin-time-picker`. Will update the PR accordingly.

PR #682, opened 2019-10-14
fix: do not prevent touchend for clear button

### LINKED ISSUE #681: Clear button doesn't work on mobile
Clicking the clear button doesn't do anything on mobile

web-padawan:
Fixes #681

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/vaadin-date-picker/682)


tomivirkki:
This should also probably validate on a clear button tap so it would behave the same on mobile and desktop. Currently, if you have a `required` date-picker, it doesn't get marked as `invalid` if you clear the field on mobile.

web-padawan:
Confirmed the regression (can be reproduced in Chrome mobile device emulation).

- desktop:`validate()` for clear button is called for empty string (correct)

- mobile: `validate()` for clear button is called for old value (incorrect).

Will continue investigating where the actual problem is.

Issue #434, opened 2017-09-14
Allow normal input from mobile


BorntraegerMarc:
### Description

The vaadin-date-picker does not allow "normal" text input from mobile. On desktop the user can click on the date picker and then just fill in the date via keyboard.

On mobile however:
![image](https://user-images.githubusercontent.com/7549295/30443832-3637c514-9981-11e7-875c-3b3d8db566a8.png)

There is no way to just let the user fill in a new date via virtual keyboard. This would be incredible handy if the date is very far away and the user does not want to scroll to that date.

### Expected outcome

A little button with the text "Input date manually". So the date picker disappears and then the user can just type in text.

### Browsers Affected

- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge
- [ ] IE 11
- [ ] iOS Safari
- [ ] Android Chrome

jouni:
Thanks for the feature request. The reason to not allow text input on touch devices is basically because the on screen keyboard would in most cases cover the entire overlay/calendar list, and we expect it to be more pleasant for users to fill in the field by browsing for the date rather than type it.

That said, I do understand that for certain use cases, like yours (selecting a date way back in the past or far in the future), the scrolling list is not ideal.

Can you elaborate your use case a bit more: do your users need to fill in dates from a very large range? I mean, could you set the initial position of the scroller to a more close date, or just use a native date input on mobile devices?

I’ve been working on a new design for vaadin-date-picker for a while now, and I will take this also into consideration if we could somehow make it easy for users to either type the date or choose from the calendar. I’m just a little afraid that iOS limitations will prevent us from doing what I envision could be an ideal solution.

BorntraegerMarc:
Basically we run a medical app, where new patients get registered. We use the date picker to choose the birthday of the patient. It's pre-filled at 1.6.1960 (because that is the medium for our new patients).

However there are still many cases where the user needs to scroll 20-40 years in the future/past.

That's why its quite important for us to being able to fill in the date manually.

To your point "the on screen keyboard would in most cases cover the entire overlay/calendar list": I envisioned it more like this:
- the user clicks on the date picker input field
- the date picker opens up. Between the button "Cancel" & "Today" there is a button "Manual". If the user clicks "Manual" the whole date picker element would disappear and basically only the text input element from step 1 would be visible. With the focus on it. Now the user can easily type any string in there.

jouni:
Alright, thanks a lot for the feedback!

[...]

PR #694, opened 2020-01-15
Add autoOpenDisabled property

niiyeboah:
When `datePicker.autoOpenDisabled = true`:
- Focusing and typing in field doesn’t open the overlay
- Clicking on icon opens the overlay
- Up / Down keys opens the overlay

### DIFF
--- a/src/vaadin-date-picker-mixin.html
+++ b/src/vaadin-date-picker-mixin.html
@@ -347,8 +347,7 @@ This program is available under Apache License Version 2.0, available at https:/
       this.addEventListener('input', this._onUserInput.bind(this));
       this.addEventListener('focus', e => this._noInput && e.target.blur());
       this.addEventListener('blur', e => {
-        this.autoOpenDisabled && this._selectAndValidateInput();
-        return !this.opened && this.validate();
+        this.autoOpenDisabled ? this._selectAndValidateInput() : !this.opened && this.validate();
       });
     }
 
--- a/src/vaadin-date-picker-mixin.html
+++ b/src/vaadin-date-picker-mixin.html
@@ -346,7 +346,10 @@ This program is available under Apache License Version 2.0, available at https:/
       this.addEventListener('keydown', this._onKeydown.bind(this));
       this.addEventListener('input', this._onUserInput.bind(this));
       this.addEventListener('focus', e => this._noInput && e.target.blur());
-      this.addEventListener('blur', e => !this.opened && this.validate());
+      this.addEventListener('blur', e => {
+        this.autoOpenDisabled && this._onOverlayClosed();
+        !this.opened && this.validate();
+      });
     }
 
     _initOverlay() {
--- a/src/vaadin-date-picker-mixin.html
+++ b/src/vaadin-date-picker-mixin.html
@@ -333,7 +333,7 @@ This program is available under Apache License Version 2.0, available at https:/
         // then the following composedPath check could be simplified down
         // to `if (!e.defaultPrevented)`.
         // https://github.com/vaadin/vaadin-text-field/issues/352
-        if (!isClearButton(e) && !this.autoOpenDisabled) {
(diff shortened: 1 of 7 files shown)


web-padawan:
@jouni do you have an idea how this should behave, keeping #434 in mind?

web-padawan:
Note, potential workaround would be to check for `_noInput`.

tomivirkki:
The feature design document we agreed upon states that "Mobile UX should stay as it is"

tomivirkki:
Ping @niiyeboah, the build is red

niiyeboah:
> If you type the value to the field and blur, it's always marked as invalid:
This should be fixed now

tomivirkki:
esc and enter buttons behave strangely with the new feature:

If you haven't yet opened the dropdown and type a value without opening the dropdown:
- ESC does nothing (should revert the changes to the input as previously)
- Enter does nothing (should apply the input's value as previously)

If you have opened and closed the dropdown earlier and type a value without opening the dropdown:
- ESC applies the input's value as the date picker's value (should revert the changes to the input as previously)
- Enter seems to work as expected

test/keyboard-input.html
```diff
@@ -700,6 +700,66 @@
...
+            expect(datepicker.value).to.equal('');
+          });
+
+          it('should apply the input value on enter when overlay not initialized', () => {
+            inputChar('1/1/2000');
+            enter();
+            expect(datepicker.value).to.equal('2000-01-01');
+          });
+
+          it('should apply input value on enter when overlay has been initialized', () => {
```
tomivirkki: This test description looks invalid (the test is about reverting the value on esc)

niiyeboah: Ah sorry my mistake, fixed now

tomivirkki:
If you input a value to a date picker and focus another one, you enter some kind of an eternal loop on IE11:
Kapture 2020-02-18 at 17 12 57

Please test the feature on all supported browsers

niiyeboah:
Rebased and refactored the implementation. Apologies for not testing all browsers 😞

src/vaadin-date-picker-mixin.html
```diff
@@ -341,7 +346,10 @@
       this.addEventListener('keydown', this._onKeydown.bind(this));
       this.addEventListener('input', this._onUserInput.bind(this));
       this.addEventListener('focus', e => this._noInput && e.target.blur());
-      this.addEventListener('blur', e => !this.opened && this.validate());
+      this.addEventListener('blur', e => {
+        this.autoOpenDisabled && this._selectAndValidateInput();
```
web-padawan: This change results in a regression: it causes `validate` to be called twice in case when date picker is closed: first time in `_selectAndValidateInput` and the second time in the next string.

Please add a test that would check that `validate` is only called once.

src/vaadin-date-picker-mixin.html
```diff
@@ -341,7 +346,10 @@
...
+        return !this.opened && this.validate();
```
web-padawan: Using `return` does nothing even if we return `false`, because [blur event](https://developer.mozilla.org/en-US/docs/Web/API/Element/blur_event) is not cancelable. 
So please remove `return` and keep the line as it was.

web-padawan:
@yuriy-fix let me request your review as you have recently changed validation logic.

src/vaadin-date-picker-mixin.html
niiyeboah: Should be fixed and added a test.

src/vaadin-date-picker-mixin.html
niiyeboah: Done

tomivirkki:
If I select a value and press esc, the input value is set as the ISO value of the component....and then blurring the field will make it invalid

dp-aod-esc

web-padawan:
@yuriy-fix @tomivirkki updated validation logic and added tests. PTAL when you have time.

PR #698, opened 2020-02-10
Optimize validation

### LINKED ISSUE #696: Validation rounds
`vaadin-date-picker` runs several validation rounds on setting the value and that leads to several small issues (i.e. `invalid` state changes firstly to `true` and then to `false` when setting the value for the first time). `change` event is dispatched before last validation round is proceeded (so the invalid state could be incorrect at that point of time).

date-p

Screenshot 2020-02-05 at 12 26 52

We should rethink the logic of the validation and get rid out of several validation rounds.

yuriy-fix:
Fixes #696


src/vaadin-date-picker-mixin.html
```diff
@@ -699,11 +699,11 @@
      * @param {string} value Value to validate. Optional, defaults to user's input value.
      * @return {boolean} True if the value is valid and sets the `invalid` flag appropriately
      */
-    validate(value) {
-      // reset invalid state on the underlying text field
-      this.invalid = false;
-      value = value !== undefined ? value : this._inputValue;
-      return !(this.invalid = !this.checkValidity(value));
+    validate() {
```
tomivirkki: Wouldn't removing the `value` argument make this a breaking change?

yuriy-fix: I guess it should not, can be considered as fix/cleanup. 
Because the validation will return the same result as previously based on the `checkValidity` `inputValid` which was checking the equality with the current `_selectedDate `. The only case when it could be different is when you manually providing the value that equals `_selectedDate` to pass the validation, but that should be achieved with custom validator / disabled validation. 🤔

src/vaadin-date-picker-mixin.html
```diff
@@ -690,7 +690,7 @@
       if (this._nativeInput && this._nativeInput.selectionStart) {
         this._nativeInput.selectionStart = this._nativeInput.selectionEnd;
       }
-      this.validate();
+      !this.value && this.validate();
```
web-padawan: Please add a comment describing why we don't have to run validation if value is set.
Has it been validated before, in `_selectedDateChanged`?

src/vaadin-date-picker-mixin.html
web-padawan: I agree that this fixes previously incorrectly working behavior.

Let's remember to mention that as "Behavior altering change" in the release notes.

src/vaadin-date-picker-mixin.html
yuriy-fix: Added.

PR #708, opened 2020-04-02
Add RTL support

### LINKED ISSUE #706: Fix keyboard navigation functionality for RTL
Switch keys (navigation) for left and right directions in the overlay.
Left should carry out right button’s functions and the opposite for right.

DiegoCardoso:
- Fix keyboard navigation when in RTL
- Add tests to cover navigation with left/right keys on RTL
- Bump versions

Fix #706

### DIFF
--- a/test/keyboard-navigation.html
+++ b/test/keyboard-navigation.html
@@ -188,7 +188,7 @@
           });
 
           ['ltr', 'rtl'].forEach(direction => {
-            const suite = describe(`[dir="${direction}"]`, () => {
+            describe(`[dir="${direction}"]`, () => {
               const isRTL = direction === 'rtl';
               before(() => {
                 document.documentElement.setAttribute('dir', direction);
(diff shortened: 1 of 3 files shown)


test/keyboard-navigation.html
```diff
@@ -187,17 +187,30 @@
...
+          ['ltr', 'rtl'].forEach(direction => {
+            describe(`[dir="${direction}"]`, () => {
+              const isRTL = direction === 'rtl';
+              before(() => {
+                document.documentElement.setAttribute('dir', direction);
+              });
+
+              after(() => {
+                if (isRTL) {
+                  document.documentElement.setAttribute('dir', 'ltr');
```
yuriy-fix, line 199: Nit: can just remove the attribute. Not a blocker.

DiegoCardoso: I'd rather force it to `ltr` because of the Safari issue I faced in context-menu.

PR #722, opened 2020-05-06
fix: keep invalid _inputValue during value-changed event

### LINKED ISSUE #721: Incorrect _inputValue on value-changed event when typing un-parseable date string
### Steps to reproduce
```javascript
datepicker.addEventListener('value-changed', e => console.log(datepicker._inputValue))
```
1. Set a date value to the date picker
2. Type "asdf" to the field and blur

**Expected outcome**: "asdf" is printed
**Actual outcome**: empty string is printed

This happens because of the following series of events:
1. `_selectedDate=null`
2. observer `_selectedDateChanged` sets `_inputValue=''` and `value=''`
3. this triggers the `value-changed` event while the `_inputValue` is incorrect
4. the old input value is added back: `_inputValue='asdf'`

This issue makes it hard to fix https://github.com/vaadin/vaadin-date-picker-flow/issues/223 without ugly workarounds.

pekam:
fix #721

### DIFF
--- a/src/vaadin-date-picker-mixin.html
+++ b/src/vaadin-date-picker-mixin.html
@@ -509,7 +509,9 @@ This program is available under Apache License Version 2.0, available at https:/
         this.__dispatchChange = true;
       }
       const value = this._formatISO(selectedDate);
-      this._applyInputValue(selectedDate);
+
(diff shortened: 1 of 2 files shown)


pekam:
Thanks!

PR #728, opened 2020-05-28
Add helper text slot and API

### LINKED ISSUE #709: Add helper text slot and API
Add slot for displaying text or component related to field.
Baseline implementation has been done for vaadin-text-field. See https://github.com/vaadin/vaadin-text-field/pull/410/
All implementations should follow the baseline implementation to achieve the same functionality.

### Requirements

- Add `helper-text` part which contains slot `helper`
- Add observer to add `has-helper` attribute when `helperText` property is set
- Add slotchange event handler to add `has-helper` attribute when content placed in helper slot
- Add helper text to `aria-describedby`
- Add observer for error message to add `has-error-message` attribute (used to hide helper in material theme when error message is set)

### Visual specifications
image

yuriy-fix:
Fixes #709 

`has-helper` and `has-error-message` are not used for styling in `date-picker`. Not sure if those should be aligned across the components, so I've not added those.


yuriy-fix:
`date-picker` is just propagating the `helperText` to the internal `text-field`. It makes sense to fix the issue in it, instead of trying to adjust it in the containing component 🤔

test/dropdown.html
```diff
@@ -97,8 +97,8 @@
       describe('alignment', () => {
 
         function assertEdgesAligned(datepickerEdge, overlayEdge) {
-          expect(datepicker.getBoundingClientRect()[datepickerEdge]).to.equal(
-            datepicker.$.overlay.$.content.getBoundingClientRect()[overlayEdge]);
+          expect(datepicker.getBoundingClientRect()[datepickerEdge]).to.be.closeTo(
+            datepicker.$.overlay.$.content.getBoundingClientRect()[overlayEdge], 0.5);
```
DiegoCardoso, line 101: Is this still needed?

PR #735, opened 2020-10-02
refactor: use tertiary theme for overlay buttons

### LINKED ISSUE #734: DatePicker-Overlay button (today, cancel) should have assigned theme tertiary
### Description
DatePicker-Overlay button (today, cancel) should have assigned theme tertiary.
Instead there's a workaround in vaadin-date-picker-overlay-content-styles.js that tries to mimick that theme.  
Unfortunately that (can) cause invisible buttons in case "background-color: transparent" is not "compatible" with default button theme.
(look for comment: /* TODO: Would be great if I could apply the "tertiary" theme from here instead of copying those styles */)

### Expected outcome
today-button / cancel-button should have assigned theme tertiary and therefore get styled accordingly

### Actual outcome
In my case today-button / cancel-button are completely transparent.

### Browsers Affected
- [x] Chrome
- [x] Firefox
- [?] Safari
- [?] Edge
- [?] IE 11
- [?] iOS Safari
- [?] Android Chrome
DatePickerOverlayTollbarButtonsAppliedTertiaryThemeManually
DatePickerOverlayTollbarButtonsMissing

DiegoCardoso:
Remove styles applied to the buttons and use "tertiary" theme instead.

Fixes #734


web-padawan:
Note, screenshots are different because of `padding` property change:

### Before

`padding: 0 0.75em;` hardcoded [here](https://github.com/vaadin/vaadin-date-picker/blob/6375bd599aa88020a06b5bb026936e3ca15287de/theme/lumo/vaadin-date-picker-overlay-content-styles.html#L139).

Screenshot 2020-10-06 at 10 03 31

### After

`padding: 0 calc(var(--lumo-button-size) / 6);` defined [here](https://github.com/vaadin/vaadin-button/blob/201faed2ad7fe82a8221431bd17eebab366d586f/theme/lumo/vaadin-button-styles.html#L129).

Screenshot 2020-10-06 at 10 03 00


I'm not sure where does `0.75em` come from. In the tertiary button theme we always used [a different padding](https://github.com/vaadin/vaadin-button/commit/d4a6944787c1daa9c4df6ebb50d044c6ba0f476f#diff-ddc18b3323f37a8861f6822bb7e37c2dR115).

web-padawan (approved):
Discussed with @jouni and agreed the change looks good.

It’s internal to our component layout, won’t affect application layouts.
If a custom theme has changed the padding, it should still work as well.

jouni:
Though, the margins should still be removed from the buttons, now that I look closer.

PR #779, opened 2020-12-07
chore: bump text-field to 3.0.0-alpha1

web-padawan:


tomivirkki (approved):

web-padawan:
There is a regression caused by `vaadin-text-field` bump, investigating.

PR #793, opened 2021-04-20
Add aria-level to divs with heading roles

semisse:
Currently, `vaadin-date-picker` has three divs with `role="heading"`. This attribute needs, concurrently, an `aria-level` reference. This is described in the [W3C WAI-ARIA 1.1 Required States and Properties](https://www.w3.org/TR/wai-aria-1.1/#heading).

To fix this a11y issue I added an `aria-level="2"`, which is the default value of this attribute.


web-padawan:
Closing in favor of https://github.com/vaadin/web-components/pull/2462. We will check that PR, sorry about the delay.

PR #809, opened 2022-10-12
fix: ensure focused date is visible if overlay is small

sissbruecker:
## Description

Fixes the overlay content to keep the focused date visible if the month scroller area does not fit a full month. Previously the overlay would only scroll to the top of the month, which means that later days in that month might not be visible. This change adds a new sub-month scroll mode to the overlay that is used when the scroll area does not fit a full month. In that mode the overlay will attempt to keep the week, that the focused date is in, visible. 

This is a best effort implementation as the scroller's position is based on months, and the scroll position of a week can not be exactly calculated. I did manual testing on Chrome, FF, Safari and IE, and it seems to work reasonably well for the default Lumo and Material themes. 

See video below for a demonstration:

https://user-images.githubusercontent.com/357820/195359489-b6dd1c58-ebea-4111-a3ea-4a34419bb95d.mov

Fixes #https://github.com/vaadin/web-components/issues/4037

## Type of change

- Bugfix


test/overlay.html
```diff
@@ -411,54 +424,55 @@
...
         beforeEach((done) => {
-          fixtureOverlayContent({
+          overlay = customizeFixture({
+            initialPosition: new Date(2021, 1, 1),
             monthScrollerItems: 0.5,
             monthScrollerOffset: 0
-          }, done);
-        });
-
-        it('should scroll when the month is above the visible area', () => {
```
sissbruecker, line 437: The old test cases for `height(visible area) < height(item)` are obsolete with the new mode.

test/overlay.html
```diff
@@ -411,54 +424,55 @@
...
+        });
+
+        it('should scroll to a position that approximately shows the week the date is in', () => {
+          // Starting on first of February
+          const initialPosition = monthScroller.position;
+          // Scroll to 15th
+          overlay.revealDate(new Date(2021, 1, 15), false);
+          const positionOf15th = monthScroller.position;
+          expect(positionOf15th).to.be.greaterThan(initialPosition);
+          expect(positionOf15th).to.be.lessThan(initialPosition + 1);
```
sissbruecker, line 447: Spent some time to create a test that checks if the date element is actually visible in the scroll area, but could not find a reliable setup. The scroller hard-codes the height of an item to 270px, while the test does not load the theme, which leads to an item only using about 135px.

So I went with the same position property checks that the other tests use.

src/vaadin-date-picker-overlay-content.html
```diff
@@ -364,7 +376,8 @@
          * Scrolls the list to the given Date.
          */
         scrollToDate(date, animate) {
-          this._scrollToPosition(this._differenceInMonths(date, this._originDate), animate);
+          const offset = this.__useSubMonthScrolling ? this._calculateWeekScrollOffset(date) : 0;
+          this._scrollToPosition(this._differenceInMonths(date, this._originDate) + offset, animate);
```
sissbruecker, line 380: This covers the initial scrolling to the focused date when opening the overlay.

sissbruecker:
Tested a bit more and found one issue. When opening the overlay for the first time, the month below the one that is currently selected briefly displays the wrong month name. In the video below the 13th of October is selected, the month below should show "November", but is briefly displayed as "October":

https://user-images.githubusercontent.com/357820/195646876-3865651c-eef5-41bf-a7f9-b144b0d2c3e9.mov

(might have to use the scrubber to find the correct frames, it's more noticable when testing live)

sissbruecker:
> When opening the overlay for the first time, the month below the one that is currently selected briefly displays the wrong month name. 

Added a commit that forces the infinite scroller to update when opening the overlay.

yuriy-fix (approved):

PR #829, opened 2023-04-24
feat: add internal has-input-value-changed event (CP: 14)

mukherjeesudebi:
## Description

Backports the following PRs:

- https://github.com/vaadin/web-components/pull/4276
- https://github.com/vaadin/web-components/pull/4335
- https://github.com/vaadin/web-components/pull/5092

Part of https://github.com/vaadin/web-components/issues/5763

### DIFF
--- a/src/vaadin-date-picker-mixin.html
+++ b/src/vaadin-date-picker-mixin.html
@@ -408,8 +408,7 @@ This program is available under Apache License Version 2.0, available at https:/
         }
       });
       this.addEventListener('keydown', this._onKeydown.bind(this));
-      this.addEventListener('input', this._onUserInput.bind(this));
-      this._inputElement.addEventListener('input', this.__onUserInput.bind(this));
+      this.addEventListener('input', this.__onUserInput.bind(this));
       this.addEventListener('focus', e => this._noInput && e.target.blur());
       this.addEventListener('blur', e => {
         if (!this.opened) {
@@ -1053,7 +1052,8 @@ This program is available under Apache License Version 2.0, available at https:/
      * @protected
      */
     _setHasInputValue(event) {
-      this._hasInputValue = event.target.value.length > 0;
+      const target = event.composedPath()[0];
+      this._hasInputValue = target.value.length > 0;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


src/vaadin-date-picker-mixin.html
```diff
@@ -1035,6 +1046,38 @@
...
+     */
+    __onUserInput(event) {
+      this._setHasInputValue(event);
+    }
+
+    /**
+      * Observer to notify about the change of private property.
+      *
+      * @private
+      */
```
vursen, line 1074:
```suggestion
    /**
     * Observer to notify about the change of private property.
     *
     * @private
     */
```

src/vaadin-date-picker-mixin.html
```diff
@@ -399,6 +409,7 @@
       });
       this.addEventListener('keydown', this._onKeydown.bind(this));
       this.addEventListener('input', this._onUserInput.bind(this));
+      this._inputElement.addEventListener('input', this.__onUserInput.bind(this));
```
vursen, line 412: Would it work if we reused the existing input listener above:

```diff
- this.addEventListener('input', this._onUserInput.bind(this));
+ this.addEventListener('input', this.__onUserInput.bind(this));
```

mukherjeesudebi: No, `should fire the event on input value presence change` this fails

vursen: I guess it is because the current `CustomEvent('input')` event we create in the test doesn't bubble although it should because the native `input` event bubbles. 

```js
dispatchEvent(new CustomEvent('input', { bubbles: true }); 
```

vursen: According to W3C specs, it should also be composed:

```js
input.dispatchEvent(new CustomEvent('input', { bubbles: true, composed: true }));
```

Let me know if it will work out.

mukherjeesudebi: No, it still doesn't

vursen (approved):

